PR #311, opened 2011-05-08
Fix a rubygems deprecation 

benlangfeld:
NOTE: Gem::SourceIndex#find_name is deprecated, use Specification.find_by_name. It will be removed on or after 2011-11-01.


benlangfeld:
For brave archeologists coming in from Google: See https://github.com/lsegal/yard/pull/1028 for followup.

PR #725, opened 2014-01-02
Rubygems 2.0+ integration

pointlessone:
Rubygems 2.0 got rid of DocManager and instead relies on pos-tinstall hook. 

This is a port of the old integration (with RDoc integration code as a basis).

There are a few caveats to this code.

Both YARD and yri data is generated into docs directories supplied by the rubygems. Currently, yri can not find generated docs there. I intend to fix it too.


lsegal:
This took forever to get around to, but it is now merged. The only problem is that `gem install` uses `ri` by default only, so real world usage of this is limited until RubyGems changes to include yri as well.

pointlessone:
Hi @lsegal!

Indeed it did take time. I don't even remember this PR. 😅

In any case, thanks and I hope it would help make YARD better.

Issue #760, opened 2014-04-04
private_class_method causes exception


nebhale:
Currently, when I run `yard (0.8.7.4)` against https://github.com/cloudfoundry/java-buildpack.git, I get a number of errors like the following:

``` bash
bhale-desktop master: bundle exece yard
[error]: Unhandled exception in YARD::Handlers::Ruby::PrivateClassMethodHandler:
[error]:   in `lib/java_buildpack/jre/memory/openjdk_memory_heuristic_factory.rb`:27:

    27: private_class_method :new
```

The documentation is built as expected, but the output isn't great.  The example class in the error can be found [here](https://github.com/cloudfoundry/java-buildpack/blob/master/lib/java_buildpack/jre/memory/openjdk_memory_heuristic_factory.rb).  If you'd like to reproduce the problem directly, you can clone the repo and run:

``` bash
bundle install
bundle exec yard
```

PR #773, opened 2014-05-10
Add helper to handle method decorators

amclain:
Expanding on the work of PR #767 for issue #760, I've added the ability for private_class_method to be handled correctly when used as a method definition decorator, which is valid syntax as of Ruby 2.1.0.


baburdick:
Any plans to merge this?

amclain:
I can update this branch if there is acceptance to it being merged in.

lsegal:
I have no objections to merging this in theory, but there is quite a lot of churn in the commits that make it hard to follow. Specifically I want to ensure that:

1. No existing functionality is being changed (this should be obvious from the tests passing)
2. No existing APIs are being changed in a backwards incompatible way
3. Most importantly, the functionality being added here should not be specific to `private_class_method` and should work with any of the existing DSL methods supported by YARD (`private`, `public`, etc). Ideally we'd want this to work for any DSL method, but that might be something that comes later.

If you can confirm these things I'm okay with taking a deeper dive on this and getting it in. I realize that the new composable "syntax" (driven by methods returning symbols) is something that a lot of developers want to begin using with YARD.

amclain:
It's been so long that I need to review the PR as well, but I believe that all three points are ok. I can also try to squash some of the commits while I'm at it.

amclain:
@lsegal I've fixed the merge conflicts and squashed the commits.

---
Summary:

This PR adds the opt-in module `YARD::Handlers::Ruby::DecoratorHandlerMethods` to assist in processing Ruby method decorators. It is used internally for the public/private visibility handlers, and can also be easily included into third-party handlers.

For example, given the decorator `foo`:

```ruby
class Test
  # The foo decorator.
  def self.foo(method)
    "foo " + new.send(method)
  end

  # The bar method.
  # @return [String] bar modified by {foo}
  foo def bar
    "bar"
  end
end
```

A handler can be created:

```ruby
class Handler < YARD::Handlers::Ruby::Base
  include YARD::Handlers::Ruby::DecoratorHandlerMethods

  handles method_call(:foo)

  def process
    process_decorator
  end
end
```

And the docstring is passed through to `#bar`:

image

`process_decorator` can also accept a block if `#bar`'s docs need to be modified based on how the decorator modifies it:

```ruby
process do
  # Set visibility of a class method to private.
  process_decorator :scope => :class do |method|
    method.visibility = :private if method.respond_to? :visibility=
  end
end
```

lib/yard/handlers/ruby/decorator_handler_methods.rb
```diff
@@ -0,0 +1,132 @@
...
+  #   # Set the decorated method's visibility to private.
+  #   process do
+  #     process_decorator :scope => :class do |method|
+  #       method.visibility = :private if method.respond_to? :visibility
+  #     end
+  #   end
+  #
+  # @!method process_decorator(*nodes, opts={}, &block)
+  def process_decorator(*nodes, &block)
+    opts = nodes.last.is_a?(Hash) ? nodes.pop : {}
```
baburdick: Would it make more sense here to use `.responds_to?(:[])` instead of `.is_a?(Hash)`? What if the implementation changes to use a Hash-like object not descended from Hash?

lsegal: That seems unlikely and would break the API in all sorts of other ways. The is-a check is sufficient, and `[]` may not be the only thing *our* implementation expects from the node argument, so validating that it's a Hash seems correct to me.

baburdick: Fair enough.

lsegal:
I finally had time to look through this PR (sorry for the delay!). In general I think it's good. I threw in some brief comments and questions inline for most of these, but I have some generalized comments:

The `process_decorators` method seems a little heavy handed in a few places. Specifically, I don't really understand the value of `transfer_source` and `transfer_docstring`. Given that users have a block form callback, they can reset any data they want if they don't want to copy over source or docstrings (i.e., revert), though these seems extremely uncommon to want to do.

More importantly though, I'd like to understand what transferring the docstring actually does. I found this to be the hardest part of the code to understand. I see that you are copying over the unparsed docstring data from the statement object, which I understand to be necessary (since the underlying method declaration will not have picked up the docstring data), but you then copy over the `tags` from the method object, which I would have suspected to be empty (since, again, the docstring data would not be associated with the method). Seeing the already-complex conditional include `method.docstring.empty?` concerns me too, because it makes me think something else is at play here.

Those were my two main issues with this. With the outstanding comments and questions resolved, I think it's getting close to being merged.

amclain:
The comments disappeared after I pushed some changes. Here's the link to the changeset with them:

https://github.com/amclain/yard/commit/534e132e3a3e99c7cf95e5ca4403193e2db3474b#commitcomment-15629753

lsegal:
For context I'm going to add follow ups to the original commit, please follow along there.

lsegal:
Added extra comments, please follow up on any of the comment threads that have not yet been responded to. Thanks!

Another high-level design point I forgot to bring up:

How does your chaining algorithm handle evaluation order of decorators? Specifically, the evaluation order with which decorators are used is important to get right. To be correct, your helper would have to handle decorators from right to left, not left to right, since Ruby evaluation would happen in RTL.

Consider two decorators, `cpub` and `cpriv`. `cpub` sets visibility to public, `cpriv` sets visibility to private (similar to `public` and `private` but chainable):

```ruby
cpub cpriv def foo; end
```

The above code should actually cause `#foo` to be public, because `cpub` actually gets evaluated last.

For YARD to be correct, it would have to emulate this order of evaluation. In my glance at the test suite, I only see one indirect test for order and it looks like it's validating LTR behavior.

To summarize: the behavior of chaining should evaluate right-to-left, and we should add explicit test cases for this scenario.

lsegal:
This took a while, but it's in. Thanks for the hard work on this, @amclain!

amclain:
Awesome, thanks @lsegal!

PR #786, opened 2014-06-12
Test for alternative class assignment

benlangfeld:
`Foo = Class.new`
is equivalent to:
`class Foo; end`

We should treat it as such, not as just any other constant.

WIP, implementation to come. Early comments welcome.


lsegal:
Marking this as closed pending real implementation. I'm not opposed to it, but I'm not in favor of it either. It really would depend on how much complexity was involved.

One difference in the calculus from my decision 2 years ago is that the legacy handlers would no longer need updating for this. You can support the regular Ruby handlers only.

PR #803, opened 2014-09-28
Add new format raw for guide template

blerins:
I would like to be able to generate new markdown files form extra files, filling includes I got from docstrings. I've created a new format `raw` for `guide` template. The idea is to keep markup, but "linkify" YARD references.

WDYT?


lsegal:
I'm going to mark this as closed and suggest publishing this as a plugin. If there is sufficient usage for this template, we could talk about pulling it in, but plugins should allow this to function normally.

PR #813, opened 2014-11-09
Fix Docstring#to_raw about @overload tag

ohai:
Using a overload tag in a module function causes a curious output with --private option. 
For example: 
    
    yard doc --private overload_in_module_function.rb

with https://gist.github.com/ohai/d720a09de10783713b82#file-overload_in_module_function-rb

This pull request fixes the problem.


lsegal:
Thanks!

Issue #692, opened 2013-08-30
New @!attribute don't work for Struct, while @attr_reader works fine


michalfita:
Following the documentation I tried to document attributes for my temporary structures by using this format:
```ruby
# @!attribute [r] something
#  @return [String] Foo goes for a drink to the Bar
B = Struct.new(:something)
```
So, the code above produces only default texts in documentation, not those I typed. While such "code" seem to be documented fine:
```ruby
# @attr_reader something [String] Foo goes somewhere else
B = Struct.new(:something)
```
Is it a bug or am I doing something wrong what does not states in the documentation clearly?

PR #820, opened 2014-12-22
Keep attribute documentation for structs.

eloyesp:
Attribute documentation using directives was overwriten by the struct handler,
lets keep that documentation.

Related to #692.


lsegal:
Thanks for this contribution, easy fix!

PR #821, opened 2014-12-26
[TASK] Configure Travis for better build performance

oliverklee:
* set sudo: false to allow Travis to use their new container-based infrastructure
* enable bundler caching
* set language: ruby for good measure


lsegal:
Awesome, thank you @oliverklee!

PR #822, opened 2014-12-29
More robust Docstring summary

tlewin:
Prevents unclosed parentheses to break the summary.
Please, see `Process.detach` at http://www.rubydoc.info/stdlib/core/Process and `IO.select` at http://www.rubydoc.info/stdlib/core/IO.


lsegal:
I like this, thanks!

PR #824, opened 2015-01-01
Remove ununsed local variables

frankywahl:
Some variables where unused. Removing them allows for less clutter and
also prevents rubyist that have warnings enabled to have those warnings
printed out.


lsegal:
This is superseded by #822, but thanks!

PR #840, opened 2015-03-03
Improved initial rendering of class list.

trevorrowe:
The sidebar `#full_list` tree of classes renders slowly and jumps after page load because much of its formatting is provided by JavaScript. Large trees suffer in render and it can be jarring when
navigating the tree. This pull request makes the following changes:

* Updates the `class_list` method in the default template to apply the initial css classes necessary to expand the tree 1 level, alternate row colors, and to indent nested items. This requires adding an additional parameter that can maintain a context of depth as this method is called recursively.

* The previous `class_list` method generated ul elements as siblings of li elements and relied on JavaScript to apply padding to simulate nesting. The ul elements for nested classes are now children of the li elements. JavaScript is no longer needed to apply nesting.

* Improves search speed on large trees by removing the use of setTimeout and instead the tree is detached from the dom and re-attached after modifications. This prevents the browser from locking up during dom updates and eliminates a large number of redraws.

Some other thoughts for improvements, not implemented by this pull-request:

* There is still a notable delay in highlighting the current "clicked" section of the tree and expanding to it upon page load. The expand logic is a result of a message sent from the frames' parent. If this message were sent upon dom load, but before the entire page load, speed would be improved here as well. Alternatively, if the frame could detect the clicked element itself based on the parents url, the delay could be drastically reduced.


trevorrowe:
Per our discussion (not in this issue), I've corrected the styling of the method and file lists. Also, I've reverted the set timeout functionality of the search and tested it on a large list to ensure there are no performance regressions.

lsegal:
Looks good, thanks @trevorrowe!

PR #843, opened 2015-03-06
Add CONSTANTSTART regular expression

hkenney:
Prior to this commit, the regular expression used in the proxy_path
method was hard coded. In addition to be inconsistent with the rest
of the regular expressions in CodeObjects::Proxy (which use the constants
defined in CodeObjects::Base), this also meant that it was not possible
to patch the regular expressions to accommodate namespaces in other
languages.

In order to fix these issues, add a CONSTANTSTART regex constant that
matches the first letter of constants that begin with an uppercase letter.

------

A bit of context on where this is coming from: I'm working on an extension for
YARD to document the Puppet DSL. However Puppet namespaces do not
require words to be capitalized. This means that our namespaces get all mangled.
I was able to trace this down to the proxy_path method.

I added another regular expression constant because patching this seemed like
a better solution that trying to patch a private method, and most of the other
regular expressions used for namespace matching have constants already. If
this is the wrong approach please let me know! Happy to try something else.

### DIFF
--- a/lib/yard/autoload.rb
+++ b/lib/yard/autoload.rb
@@ -46,6 +46,7 @@ module YARD
     autoload :BUILTIN_MODULES,      __p('code_objects/base')
     autoload :BUILTIN_EXCEPTIONS,   __p('code_objects/base')
     autoload :CONSTANTMATCH,        __p('code_objects/base')
+    autoload :CONSTANTSTART,        __p('code_objects/base')
     autoload :METHODMATCH,          __p('code_objects/base')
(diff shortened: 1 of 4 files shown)


lsegal:
Looks good thanks!

hkenney:
Thanks so much @lsegal! 

Do you happen to have an idea of when you'll be releasing the next version of the YARD gem? Just trying to plan out if we should have a temporary workaround or just wait if it's going to be in the next month or so.

PR #860, opened 2015-04-29
[config_dir] use one constant and fallback to /tmp

ChrisLundquist:
This lets applications launch when ENV["HOME"] is unset, blank, or
relative.

For a variety of reasons, when launching my app, HOME is empty
and gems I need depend on YARD. I believe we shouldn't prevent
apps from starting in this case.

This cleans up our usage of "~/.yard" in a few places to reference the
same constant, which falls back to "/tmp.yard" when HOME is unset.

This also removes the deprecated ::YARD::CONFIG_DIR.
It was deprecated [Sep 16, 2010
598b051202549122dc12411ce6e7b26b4c98d66b](https://github.com/lsegal/yard/commit/598b051202549122dc12411ce6e7b26b4c98d66b)

### DIFF
--- a/lib/yard.rb
+++ b/lib/yard.rb
@@ -7,9 +7,6 @@ module YARD
   # The root path for YARD builtin templates
   TEMPLATE_ROOT = File.join(ROOT, '..', 'templates')
 
-  # @deprecated Use {Config::CONFIG_DIR}
-  CONFIG_DIR = File.expand_path('~/.yard')
-
   # An alias to {Parser::SourceParser}'s parsing method
   #
   # @example Parse a glob of files
--- a/lib/yard.rb
+++ b/lib/yard.rb
@@ -1,4 +1,5 @@
 require File.expand_path('../yard/version.rb', __FILE__)
+require File.expand_path('../yard/config.rb', __FILE__)
 
 module YARD
   # The root path for YARD source libraries
@@ -6,6 +7,9 @@ module YARD
 
   # The root path for YARD builtin templates
   TEMPLATE_ROOT = File.join(ROOT, '..', 'templates')
+
+  # @deprecated Use {Config::CONFIG_DIR}
+  CONFIG_DIR = ::YARD::Config::CONFIG_DIR
 
   # An alias to {Parser::SourceParser}'s parsing method
   #
--- a/lib/yard/cli/yri.rb
+++ b/lib/yard/cli/yri.rb
@@ -6,12 +6,12 @@ module YARD
     class YRI < Command
       # The location in {YARD::CONFIG_DIR} where the YRI cache file is loaded
       # from.
-      CACHE_FILE = File.expand_path('~/.yard/yri_cache')
+      CACHE_FILE = File.join(::YARD::Config::CONFIG_DIR, 'yri_cache')
 
       # A file containing all paths, delimited by newlines, to search for
       # yardoc databases.
       # @since 0.5.1
-      SEARCH_PATHS_FILE = File.expand_path('~/.yard/yri_search_paths')
+      SEARCH_PATHS_FILE = File.join(::YARD::Config::CONFIG_DIR, 'yri_search_paths')
 
       # Default search paths that should be loaded dynamically into YRI. These paths
       # take precedence over all other paths ({SEARCH_PATHS_FILE} and RubyGems
--- a/lib/yard/config.rb
+++ b/lib/yard/config.rb
@@ -90,10 +90,9 @@ module YARD
       attr_accessor :options
     end
 
-    # The location where YARD stores user-specific settings
 
     HOME_DIR = begin
-                 Dir.home
+                 File.expand_path('~')
                rescue ArgumentError => e
                  require 'tmpdir'
                  dir = Dir.tmpdir
@@ -103,6 +102,7 @@ module YARD
                  STDERR.puts("Falling back to temp directory '#{dir}' for HOME.")
                  dir
                end
+    # The location where YARD stores user-specific settings
     CONFIG_DIR = File.join(HOME_DIR, '.yard')
 
     # The main configuration YAML file.
--- a/lib/yard/config.rb
+++ b/lib/yard/config.rb
@@ -1,3 +1,4 @@
+require 'etc'
 module YARD
   # This class maintains all system-wide configuration for YARD and handles
   # the loading of plugins. To access options call {options}, and to load
@@ -91,7 +92,7 @@ module YARD
     end
 
     # The location where YARD stores user-specific settings
-    CONFIG_DIR = File.expand_path('~/.yard') rescue '/tmp/.yard'
+    CONFIG_DIR = File.expand_path('~/.yard') rescue Etc.getpwuid.dir
 
     # The main configuration YAML file.
     CONFIG_FILE = File.join(CONFIG_DIR, 'config')
--- a/lib/yard/config.rb
+++ b/lib/yard/config.rb
@@ -91,7 +91,7 @@ module YARD
     end
 
     # The location where YARD stores user-specific settings
-    CONFIG_DIR = File.expand_path('~/.yard')
+    CONFIG_DIR = File.expand_path('~/.yard') rescue '/tmp/.yard'
 
     # The main configuration YAML file.
     CONFIG_FILE = File.join(CONFIG_DIR, 'config')
--- a/lib/yard/config.rb
+++ b/lib/yard/config.rb
@@ -1,4 +1,3 @@
-require 'etc'
 module YARD
   # This class maintains all system-wide configuration for YARD and handles
   # the loading of plugins. To access options call {options}, and to load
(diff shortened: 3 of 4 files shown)


lsegal:
I'm not sure I like the idea of pointing to /tmp. A hard error seems more reasonable than silently pointing to a place on the filesystem that the user does not know about. If a user was using YARD to write configuration to disk, that configuration would get lost. 

If I were nitpicking even more, I'd say /tmp doesn't even guarantee that no error gets thrown, since Windows doesn't have /tmp and many OS's don't *require* that path to be present, so ultimately this wouldn't solve the error of YARD being unable to start up in the case of <insert environment-specific scenario here>. Basically this patch is just replacing one dependency (a common environment variable) with another (a common path).

> This also removes the deprecated ::YARD::CONFIG_DIR. It was deprecated Sep 16, 2010 598b051202549122dc12411ce6e7b26b4c98d66b

I would prefer if this was not removed.

That said-- why not just set HOME when running YARD? Seems easier than relying on /tmp.

ChrisLundquist:
> Windows doesn't have /tmp and many OS's don't require that path to be present,

That's fair. I would be happy to fall back to some other path.

> I would prefer if this was not removed.

I'd be happy to add it back. The reason I removed it was to point all our derived constants to the same base constant. That particular file didn't seem to be able to reference `::YARD::Config...`


The issue isn't with running yard, it is with running unicorn as nobody.
Yard is stopping my app from booting because nobody has no home.
I tried giving nobody a home, but then rvmsudo removes HOME from ENV.
Pry depends on yard, which is why yard gets pulled in.

In my unicorn.rb I can do something like

```ruby
before_exec do |server|
  ENV["HOME"] = "/tmp"
end
```

But that feels like a poor solution.
This is a system I recently inherited, and there are a lot of things that could be done better. This is how it is at the moment though, and this was the most reasonable solution I could find at the moment.

ChrisLundquist:
http://stackoverflow.com/a/13670703 suggests that
```ruby
require 'etc'
Etc.getpwuid.dir
```
Might be a decent fallback to get a home directory.

ChrisLundquist:
I have made the changes you requested and think I've found a middle ground. Let me know if there is anything else that I can do.

JackDanger:
@lsegal My team just found this PR after frustratingly bouncing against this error. How about merging and releasing a new version?

lsegal:
@JackDanger the easier fix is here is to use a workaround that defines HOME for your environment. A few really simple workarounds are shown above. 

Fixing this in YARD doesn't actually solve the underlying problem in your environment, namely that *Ruby*, not YARD, requires a HOME to be defined in order for `File.expand_path()` to work. YARD might be one of your tools that calls `File.expand_path`, but I doubt that it will be ever be the only library that does. If you continue to run without a HOME defined you will eventually run into another library that trips up your application for the same reason.

As for merging this in, this patch, in its current state, is not merge-able.

1. It's not cross-platform. It relies on 'etc' which may not be available on Windows systems or other Ruby implementations.
2. The actual behavior is that it will rescue and return `Etc.getpwuid.dir` directly without tacking on the necessary `.yard` path. This proposed change would not behave properly and introduce a regression.
3. Blind `rescue` calls are an anti-pattern. If File.expand_path failed for any other reason it would be hidden.
4. And also, not really discussed much, but `File.expand_path` will still fail anywhere else we might decide to use it in YARD. And we use it in quite a few places: https://github.com/lsegal/yard/search?utf8=%E2%9C%93&q=expand_path -- YARD might "load" with this patch in place, but it probably still won't work correctly unless we stopped using `File.expand_path` completely.

/cc @ChrisLundquist

ChrisLundquist:
I'd be happy to make any changes required to resolve this issue. I agree, this is a work around for other issues. Wouldn't it be wonderful to make Yard a little more flexible for these situations?

> It's not cross-platform. It relies on 'etc' which may not be available on Windows systems or other Ruby implementations.

On my windows machine I can do the following:
```ruby
require 'tmpdir'
Dir.tmpdir
=> "C:/Users/<user>/AppData/Local/Temp"
```

This isn't quite as well behaved on OS X
```ruby
requiirb(main):001:0> require 'tmpdir'
=> true
irb(main):002:0> Dir.tmpdir
=> "/var/folders/f9/62lkdjbs0jq23ngdyc2866rh0000gn/T"
```

I'm happy to make this more cross platform if this is the main sticking point.


> The actual behavior is that it will rescue and return Etc.getpwuid.dir directly without tacking on the necessary .yard path. This proposed change would not behave properly and introduce a regression.

You're right. That's my mistake. We can fix that up with something like:
```Ruby
irb(main):002:0> File.join(File.expand_path("~"), ".yard")
=> "/Users/clundquist/.yard"
```


> Blind rescue calls are an anti-pattern. If File.expand_path failed for any other reason, the error would be hidden.

That's true. We can make this rescue more specific. I was optimizing for minimal changes to get the conversation going.

> And also, not really discussed much, but File.expand_path will still fail anywhere else we might decide to use it in YARD. And we use it in quite a few places: https://github.com/lsegal/yard/search?utf8=%E2%9C%93&q=expand_path -- YARD might "load" with this patch in place, but it probably still won't work correctly unless we stopped using File.expand_path completely. ALL of your own application code would have to ensure the same.

`File#expand_path` only breaks when using `~`, it resolves relatives paths and symlinks just fine. This is why I took the approach of trying to only use `~` in one place.

JackDanger:
> YARD might be one of your tools that calls File.expand_path, but I doubt that it will be ever be the only library that does

It's the only one that calls it in this way. I added YARD to my team's Gemfile and, while it passed the build, it failed to boot at deploy time. Everybody was both confused about what I'd done and then, when we realized what was going on, stunned that adding YARD to an existing real Rails app introduced this behavior. I @ChrisLundquist is on the right track here to keep YARD's source quality high and also fix this for YARD's users.

lsegal:
> It's the only one that calls it in this way. 

For now. It sounds like you should be fixing the root cause of your issue, not the symptom. To be clear, YARD is the symptom, not your problem. At some point in the future, you might run into this again, and your new developers will be just as confused. Stop stunning your developers and fix the problem at its core. Create a robust deployment environment that does not fail due to extremely simple and standard Ruby operations. In other words, if your application is crashing because of a simple line like `File.expand_path('~')`, then the problem is in your application, not YARD.

That said, I agree that if we can help to avoid obvious mishaps we should. The key here is finding the *right* fix, and we haven't done that yet, which is why this PR has not been merged yet.

As a sidenote, I'm not entirely sure what it is your Rails application does, but if you're not actually using YARD in the runtime of your Rails app, you really should *not* be loading it into your runtime. I could be wrong, but based on the "I added YARD to my team's Gemfile" comment I take it you probably wanted to add it to your `:development` group, not `:production`. If you actually are using YARD in your app I'd love to know what you're doing with it, out of curiosity!

ChrisLundquist:
> That said, I agree that if we can help to avoid obvious mishaps we should. The key here is finding the right fix, and we haven't done that yet, which is why this PR has not been merged yet

I'd really like to find a way that accomplishes this. I outlined a few ways forward above. If those changes would bring this PR up to par, I'm happy to do them. My only hesitation is wasted work.


> As a sidenote, I'm not entirely sure what it is your Rails application does, but if you're not actually using YARD in the runtime of your Rails app, you really should not be loading it into your runtime

In my case, occasionally our developers need to poke at things in production / staging and like to use pry. Pry requires Yard, and then things break.

lsegal:
I honestly think the right approach here would be to:

1. Define a `YARD::HOMEDIR` that did a lookup of `ENV['HOME'] || ENV['HOMEPATH']` in the nature of http://stackoverflow.com/questions/4190930/cross-platform-means-of-getting-users-home-directory-in-ruby

2. Replace all use of File.expand_path constants with `File.join(YARD::HOMEDIR, '.yard')` etc.

ChrisLundquist:
Bash steps to create a user along the lines of what we have:
```
sudo useradd foo
sudo su - foo
# sudo su - foo prints a warning and sets HOME to /
# No directory, logging in with HOME=/
unset HOME 
irb
```

```ruby
2.2.0 :001 > Dir.home
ArgumentError: couldn't find HOME environment -- expanding `~'
	from (irb):1:in `home'
	from (irb):1
	from /usr/local/rvm/rubies/ruby-2.2.0/bin/irb:11:in `<main>'
2.2.0 :002 > ENV['HOMEPATH']
 => nil 
2.2.0 :003 > ENV['HOME']
 => nil
```

This is why I went with a temp directory.

lsegal:
Sure, a fallback would be fine on the HOMEDIR constant.

ChrisLundquist:
I gave this another shot.

bf4:
FWIW, in [SimpleCov we did this](https://github.com/colszowka/simplecov/blob/3e8f3e372cd8b4d51816d71f51b40711b6cdac10/lib/simplecov/defaults.rb#L90-L96) to work around doing our best to find the home dir. If we couldn't we didn't try to use it.

```ruby
# Autoload config from ~/.simplecov if present
require "etc"
home_dir = File.expand_path("~") || Etc.getpwuid.dir || (ENV["USER"] && File.expand_path("~#{ENV['USER']}"))
if home_dir
  global_config_path = File.join(home_dir, ".simplecov")
  load global_config_path if File.exist?(global_config_path)
end

```

lsegal:
I'm deciding to close this. I did add in refactoring of ~/.yard, but I cannot take in a change that will silently point to /tmp for configuration, because I don't see how that would benefit the user in any way. It should be trivial for any app to meet this minimum bar of ensuring HOME is set. 

If HOME is not defined, YARD should probably hard fail. I'd accept a patch that failed early, but not one that launched when configuration would not be properly sourced or saved. It doesn't seem like such a patch is being proposed in this PR though.

PR #865, opened 2015-05-14
Drop old Rubies from .travis.yml

oliverklee:
Support for Ruby 1.9.3 has ended on 2015-02-23. So we can drop all Rubies
1.9.3 and lower from .travis.yml.


lsegal:
YARD still supports these versions, so as long as Travis will run the tests, we will continue to run them.

PR #878, opened 2015-06-12
Specify required Ruby version

nicolasleger:


lsegal:
What does this fix?

nicolasleger:
To specify the Ruby version for people as used by Rubygems for example at https://rubygems.org/gems/yard

lsegal:
I'm not sure that this actually fixes an existing problem besides adding more metadata. Unfortunately this has side effects that would affect runtime behavior, so it's not just an aesthetic change.

lsegal:
Marking as WontFix. This change introduces an unnecessary new "yard" file and increases the minimum version requirement.

PR #882, opened 2015-06-23
Use --no-document option instead of --no-rdoc and --no-ri option

takiy33:
--no-rdoc and --no-ri option is obsolated.
Thank you.


lsegal:
Seems good thanks!

PR #897, opened 2015-09-18
Extract method parameter names from C function signatures

alexdowad:
Use a couple of regexes on the source for a Ruby method defined in a C extension, to get the parameters. This is already done for methods defined in Ruby, but for those defined in C.

This is not foolproof -- for example, if the Ruby method implemented by a C function takes a block argument, we can't identify that from the C function signature. But at least we can identify regular method arguments.

This makes the generated documentation for C extension methods noticeably better.

### DIFF
--- a/lib/yard/handlers/c/handler_methods.rb
+++ b/lib/yard/handlers/c/handler_methods.rb
@@ -147,6 +147,7 @@ module YARD
           if src_stmt = symbols[symbol]
             register_file_info(object, src_stmt.file, src_stmt.line, true)
             register_source(object, src_stmt)
+            record_parameters(object, symbol, src_stmt)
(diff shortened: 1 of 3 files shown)


alexdowad:
Just adjusted the regexes to work in Ruby 1.8.7.

alexdowad:
Just added another little tweak (cleanup of a regex) to this PR... please have a look.

alexdowad:
I removed the regex tweak as requested.

alexdowad:
Ping...

lsegal:
Much better, thanks @alexdowad!

alexdowad:
Thanks to you for providing such a great tool, and please let me know how I can make my other PRs better as well.

PR #898, opened 2015-09-18
Remove unused variable

iankronquist:
It looks like this was left over after a refactor in 2010. It's funny what
slips into code sometimes.


lsegal:
Thanks!

PR #899, opened 2015-09-21
Recognize core Ruby classes (declared in ruby.h) in C extension Init function

alexdowad:
In a C extension's Init method, code like this:

    rb_define_method(rb_cFixnum, "popcount", fix_popcount, 0);

...doesn't cause Fixnum#popcount to appear in the generated Yardocs, because YARD
can't find a definition of Fixnum anywhere in the C extension code.

Fix that, so that when a C extension adds a method to a core Ruby class or module,
the method documentation will appear in the generated Yardocs.

### DIFF
--- a/lib/yard/handlers/c/handler_methods.rb
+++ b/lib/yard/handlers/c/handler_methods.rb
@@ -51,6 +51,16 @@ module YARD
           end
 
           namespace = namespace_for_variable(var_name)
+
+          # Is this method being defined on a core Ruby class or module?
+          if namespace.is_a?(Proxy)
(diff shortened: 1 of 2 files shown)


alexdowad:
Just fixed a typo.

alexdowad:
OK, now it only registers a class or module when it sees a method definition (as per your comment).

lsegal:
Thanks for all three patches @alexdowad!

alexdowad:
Thank you! Question: would you accept patches which update your test suite to use RSpec 3 rather than RSpec 2?

lsegal:
@alexdowad sure, the codebase is stable enough at this point to do that refactor without stepping on toes.

PR #900, opened 2015-09-21
Stricter matching of override comments to methods defined in C extension

alexdowad:
Previously, an override comment for `Foo#baz` would also be applied to `Bar#baz`.
This was especially problematic, because if an extension defined methods with the
same names on two different classes, the override comment which appeared *first*
in the source code would always be favored over a more specific comment which
appeared later.

One thing I'd like to clarify:

If there is an override comment for `Foo::foo`, should it be attached to *both* `Foo.foo` *and* `Foo#foo`?

### DIFF
--- a/lib/yard/handlers/c/handler_methods.rb
+++ b/lib/yard/handlers/c/handler_methods.rb
@@ -158,10 +158,15 @@ module YARD
           # so look in overrides
           override_comments.each do |name, override_comment|
             next unless override_comment.file == file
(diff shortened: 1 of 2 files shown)


alexdowad:
Just updated this PR so it applies cleanly onto `master`. Please have a look.

alexdowad:
Ping...

lsegal:
This looks good sorry about the delay in getting to look into this!

PR #904, opened 2015-10-12
Update style.css - code windows tab-size set to 2

MSP-Greg:
Line 450, changes pre.code style, setting tab-size to 2 rather than browser default (often 4 or5)


lsegal:
Thanks!

PR #907, opened 2015-10-23
Upgrade RSpec to 3.3

alexdowad:
Let's just make sure all the tests pass on CI...

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,2 @@
+--color
+--format documentation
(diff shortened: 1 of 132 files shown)


alexdowad:
I investigated the failure on MRI 1.8.7 -- it is related to the order which objects appear in `Hash#values`. In 1.8, Ruby hashes did not have a defined iteration order. Now they do.

lsegal:
Thanks for all this awesome work! I'll give it an extra set of eyes to make sure everything translated properly.

PR #919, opened 2015-12-03
Seems to fix Chrome method signature height issue

MSP-Greg:
Or, the extra tall 'blueish' method signatures.

Not sure about where p.signature elements are used.  Used 2.2 simply based on twice
font-size...

### DIFF
--- a/templates/default/fulldoc/html/css/style.css
+++ b/templates/default/fulldoc/html/css/style.css
@@ -245,6 +245,7 @@ p.signature, h3.signature {
   font-size: 1.1em; font-weight: normal; font-family: Monaco, Consolas, Courier, monospace;
   padding: 6px 10px; margin-top: 1em;
   background: #E8F4FF; border: 1px solid #d8d8e5; border-radius: 5px;
+  line-height:2.2em;
 }
 p.signature tt,
 h3.signature tt { font-family: Monaco, Consolas, Courier, monospace; }


MSP-Greg:
Loren,

Used at http://msp-greg.github.io/su_info/index.html.

One test issue - on RubyDoc.info, there are the floated 'permalink' elements.  I can't check html with those easily.

lsegal:
Permalinks come from the `yard server`, you can test there. I'm trying to understand what the actual issue is here. I see the larger height, but I don't see this reproduced when generating HTML here. Can you list the Ruby version / OS / YARD version, etc., you used to generate the output with?

MSP-Greg:
Sorry, I'm in the middle of some other coding... The issue is only about how some elements render in Chrome vs other browsers.  Images grabbed from RubyDoc.info.  Firefox seems to have a bit larger fonts...

In the two below images, notice the difference in height of the signature elements (light blue backgrounds).  On an iPad, display matches the Firefox styling.

Small item.  Fix makes Chrome render like Firefox / iPad, and doesn't seem to affect Firefox, iPad.

Greg

**Chrome screen grab**
---

yard_chrome

**Firefox screen grab**
---
yard_firefox

lsegal:
I guess the confusing part for me is that Chrome renders the same page fine on my machine:

rdoc

What OS / Chrome version are you running against?

MSP-Greg:
Well, I'm more confused.  I was working on a Vista system (don't ask), but it has the current version of Chrome (47.0.2526.73) and Firefox.  I checked it on a) iPad, b) Android tablet (both Chrome & standard browser), and c) Windows 10, which has the same version of Chrome as the Vista system, and Edge.

The **only browser** with the tall sections is Chrome on the Vista system.  Hence, never mind...

Greg

lsegal:
I'm not opposed to fixing this, just wanted to understand the root of the issue. I'll check into this and possibly merge if it doesn't cause other issues.

MSP-Greg:
'Fixing this' implies something wrong with YARD.  Now that I've investigated further, I'm not too concerned.  I apologize for assuming that the same version of Chrome would act the same on all desktop platforms...

lsegal:
Closing this, then. Thanks for investigating! :)

MSP-Greg:
Found this on Slashdot, some comments are as recent as fall '14.  Due to Chrome's user agent stylesheet of some block tags.  Chrome does the same thing with H1, H2, & H3, but what's needed to 'override' is unclear and works differently on (at least) H3.

http://stackoverflow.com/questions/5721966/webkit-margin-adds-unwanted-margin-on-text

Happy Holidays

PR #923, opened 2015-12-18
Image assets not rendering in server on windows

MSP-Greg:
Looked into 'asset' images that display fine in doc but not server.  Also, on windows.

Single line change seems to solve the problem

Happy Holidays!


MSP-Greg:
Okay, now that I took break...

Replace with

render IO.binread(filename)

MSP-Greg:
Cool.  On windows, using server

render File.read(filename) does not work.  Replacing with

render File.read_binary(filename)  does work.  New pull request?

lsegal:
You should just be able to update this one.

lsegal:
Thanks!

PR #924, opened 2015-12-19
Redcarpet - underscores

MSP-Greg:
Correct option is
:no_intra_emphasis
not
:no_intraemphasis

See
http://www.rubydoc.info/github/vmg/redcarpet/master


lsegal:
`:no_intraemphasis` should be part of backcompat support: https://github.com/vmg/redcarpet/blob/master/lib/redcarpet/compat.rb#L29

MSP-Greg:
Maybe I've been at this too long today.  Further checking shows that if the underscore is preceded by something other than [A-Za-z0-9] (I was using '?') things need escaping.  A Redcarpet quirk.  My mistake.

Greg

lsegal:
No problem! I was of the opinion that Redcarpet should be documenting their backwards compatibility layer, which would have clarified.

PR #926, opened 2015-12-24
Minor text correction

ufo2mstar:
just a minor text correction 
contents => contains


lsegal:
The example is actually correct, the first word is the name of the parameter being documented.

Issue #929, opened 2016-01-08
plugins are not automatically detected when running yardoc


bertBruynooghe:
Steps to reproduce in 0.8.7.6:

1. use some unknown tag in documentation of class => error both when running ```yardoc``` and ```yard server -r```
2. create and install a gem named "yard-..." which defines the tag => still error in ```yardoc```,  ```yard server -r``` seems to pick up the tag.

PR #930, opened 2016-01-08
documented autoloading plugins for yardoc usage

bertBruynooghe:
to solve ticket #929:
`yard config load_plugins true`


bertBruynooghe:
Closed PR; opened nes PR https://github.com/lsegal/yard/pull/934 instead.

lsegal:
I decided to rework the majority of the section since a lot of it was heavily outdated anyway (it actually suggested to install a gem using gemcutter.org as a source!). I removed the notes about versions, since most of this is no longer relevant.

Thanks for bringing this old doc to my attention though!

Issue #888, opened 2015-07-16
ArgumentError in VisibilityHandler on private method with gsub


akostadinov:
Using ruby 2.2.2 and yard 0.8.7.6

Here's my method:
```
    private def ssh_opts(opts)
      ssh_opts = {}
      properties.each { |prop, val|
        if prop.to_s.start_with? 'ssh_'
          ssh_opts[prop.to_s.gsub(/^ssh_/,'').to_sym] = val
        elsif prop == :user
          ssh_opts[:user] = val
        end
      }
      ssh_opts.merge! opts

      return ssh_opts
    end
```

If I remove the gsub line, then all works fine. But in the way it is written above, I'm getting:
```
[error]: Unhandled exception in YARD::Handlers::Ruby::VisibilityHandler:
[error]:   in `lib/host.rb`:406:

	406: private def ssh_opts(opts)

[error]: ArgumentError: invalid empty object name
[error]: Stack trace:
	/home/avalon/.gem/ruby/gems/yard-0.8.7.6/lib/yard/code_objects/base.rb:168:in `new'
	/home/avalon/.gem/ruby/gems/yard-0.8.7.6/lib/yard/handlers/ruby/visibility_handler.rb:20:in `block (2 levels) in <class:VisibilityHandler>'
	/home/avalon/.gem/ruby/gems/yard-0.8.7.6/lib/yard/parser/ruby/ast_node.rb:209:in `block in traverse'
	/home/avalon/.gem/ruby/gems/yard-0.8.7.6/lib/yard/parser/ruby/ast_node.rb:208:in `each'
	/home/avalon/.gem/ruby/gems/yard-0.8.7.6/lib/yard/parser/ruby/ast_node.rb:208:in `with_index'
	/home/avalon/.gem/ruby/gems/yard-0.8.7.6/lib/yard/parser/ruby/ast_node.rb:208:in `traverse'
```

Couldn't find out why is it like that and couldn't find any workarounds.

lsegal:
> Couldn't find out why is it like that and couldn't find any workarounds.

Because this syntax is not yet supported in YARD. Marking this as a feature request.

akostadinov:
The interesting thing is that if I remove `gsub(/^ssh_/,'')`, not the whole line, I don't see any such errors. If I remove the `private` keyword, I don't see such error. I mean, why should `gsub` have anything to do with it? I have a lot of `private def ...` definitions.

lsegal:
[...]

PR #933, opened 2016-01-14
Handle 2.1-style visibility where the def statement returns a symbol.

mattparlane:
This is an attempt at fixing #888. Please let me know what you think. Thanks!


mattparlane:
Apologies, I didn't change the legacy parser and I see the test suite is failing. Perhaps if you could let me know if I'm on the right track? I'm not even sure how the feature should work in legacy mode since the syntax isn't supported... thoughts?

mattparlane:
I've just noticed #773, which looks like it does the same job but much, much better. Please consider this a vote for getting that PR merged, and feel free to close this one. Thanks!

PR #946, opened 2016-03-09
Use last_description instead of last_comment for Rake 11

hsbt:
Rake 11 removed `last_comment`. It should be replace `last_description`

see. https://github.com/ruby/rake/blob/master/History.rdoc

### DIFF
--- a/lib/yard/rake/yardoc_task.rb
+++ b/lib/yard/rake/yardoc_task.rb
@@ -66,7 +66,7 @@ module YARD
       # Defines the rake task
       # @return [void]
       def define
-        desc "Generate YARD Documentation" unless ::Rake.application.last_comment
+        desc "Generate YARD Documentation" unless ::Rake.application.last_description
         task(name) do
           before.call if before.is_a?(Proc)
           yardoc = YARD::CLI::Yardoc.new


jodosha:
This is a blocking issue. Please consider to merge and release a patch version. Thank you. :+1:

lsegal:
@jodosha this shouldn't be blocking unless you absolutely need Rake 11, which looking at the ChangeLog, shouldn't be the case. Locking to `rake ~> 10.0` should be a reasonable workaround.

jodosha:
@lsegal Thanks for the quick reply :tophat:. Sure, locking Rake works, but I'm sure that there are a lot of developers out there with the same problem.

Instead of let the rest of us to lock Rake to `~> 10.0`, it would be nice if you can make it to work with latest Rake.

All the best. And once again, thanks for YARD. :green_heart:

bbatsov:
Seems this was deprecated a while back, so the patch should be safe.

jnbt:
CI is failing as Rake 11 doesn't support Ruby 1.8.7 anymore.

jodosha:
@lsegal Thanks for merging this 💚 .

PR #953, opened 2016-03-28
RackAdapter should understand if it's serving under a script_name

ento:
Demo: https://github.com/ento/mountable-yard-rack-adapter-example

## Motivation

Current implementation of the server assumes it's serving under the root path (`/`). This prohibits mounting `YARD::Server::RackAdapter` under a prefix path like:

```
run Rack::URLMap.new("/api" => yard_adapter)
```

## Changes

* Use `request.path_info` instead of `request.path` where appropriate
* <s>Renamed `DocServerHelper#base_path` to `#lib_base_path` and have it return an absolute path</s>
* Added `DocServerHelper#site_abs_path`
* All server spec runs under a mock request with `script_name` populated


lsegal:
Thanks for the PR! 

First, this seems to [break existing tests](https://travis-ci.org/lsegal/yard/builds/118905122), and also introduces a breaking change. I can consider this only if tests are passing and no breaking changes are introduced.

At a quick glance, I'm curious why `path_info` vs `path` is used, especially when it seems as though you are adding the `script_name` back to various URLs. Can you explain a bit about what the problem with the existing implementation is?

ento:
> Can you explain a bit about what the problem with the existing implementation is?

I wanted to make YARD docs available under `/dev/yard` in a Rails development environment, but ran into 404 errors.

This is because the current implementation tries to resolve the route using `request.path`, which includes the `/dev/yard` part:

```
 /--- path ----\
 /dev/yard/docs/
 \-------/\----/
script_name path_info
```

The same goes for resolving static asset paths. The server needs to strip out the `/dev/yard` part before trying to match with a file path.

I think the right approach is to have `YARD::Server` only concern itself with `request.path_info` (hence `s/path/path_info/g`), and worry about generating the correct absolute URL path when needed. That was the intent of this PR; I'll see if I can make it backwards-compatible.

lib/yard/server/templates/default/layout/html/script_setup.erb
```diff
@@ -1,7 +1,7 @@
 <script type="text/javascript" charset="utf-8">
   pathId = <%= @path ? @path.inspect : 'null' %>
   relpath = '/';
-  docsPrefix = '<%= base_path(router.docs_prefix) %>';
-  listPrefix = '<%= base_path(router.list_prefix) %>';
-  searchPrefix = '<%= base_path(router.search_prefix) %>';
+  docsPrefix = '<%= site_abs_path(base_path(router.docs_prefix)) %>';
+  listPrefix = '<%= site_abs_path(base_path(router.list_prefix)) %>';
+  searchPrefix = '<%= site_abs_path(base_path(router.search_prefix)) %>';
```
ento: I'm not sure if these prefixes should be absolute paths. What are they used for?

ento:
Now tests are passing (thanks for fixing the 1.8 issue), and doesn't touch the `base_path` helper. Appreciate further feedbacks.

lsegal:
Thanks for the PR! I had to make some adjustments (not all `path -> path_info` changes were technically correct, missing explicit test cases for new functionality), but it should be in the next release coming soon.

ento:
Thanks!

PR #954, opened 2016-03-29
Make the C Handler process `rb_global_const`. Fixes #915.

### LINKED ISSUE #915: rb_define_global_const not documented?
Global constants defined in `.rb` files appear to be documented:
```
# This is a test.
HELLO_RUBY = 'Hi!'.freeze
```

That will appear under "Top Level Namespace".

But defining a global constant in C doesn't seem to document:
```
// Hello C World.
rb_define_global_const("SOME_NAME", Qtrue);
```

That doesn't appear anywhere in the docs - even though the constant regex appear to take into account the pattern for this function.

Is this intentional somehow or a bug?
If it's intentional, how can a plugin add constants to the Top Level Namespace?

thomthom:
I'm not that familiar with the exact code style and structure, so let me know if it needs tweaks.

### DIFF
--- a/lib/yard/handlers/c/handler_methods.rb
+++ b/lib/yard/handlers/c/handler_methods.rb
@@ -103,8 +103,12 @@ module YARD
         end
 
         def handle_constants(type, var_name, const_name, value)
-          return unless type == 'const'
-          namespace = namespace_for_variable(var_name)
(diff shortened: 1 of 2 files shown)


lsegal:
Looks good to me!

PR #967, opened 2016-06-08
Test doesn't need to be in loop

MSP-Greg:


lsegal:
Thanks! 

The JRuby error looks unrelated.

Issue #902, opened 2015-10-05
Documenting private constants fails when `private_constant` argument is not a symbol


backus:
Here are a few examples

```sh
$ cat example1.rb
class Foo
  BAR, BAZ, BUX = 1, 2, 3

  # constants(false) # => [:BAR, :BAZ, :BUX]
  private_constant(*constants(false))
end
$ cat example2.rb
class Foo
  BAR = 1

  def self.private_constant_symbol
    :BAR
  end

  private_constant(private_constant_symbol)
end
$ cat example3.rb
class Foo
  BAR = 1

  @constant_name = :BAR

  private_constant(@constant_name)
end
$ yardoc example1.rb
[error]: Unhandled exception in YARD::Handlers::Ruby::PrivateConstantHandler:
[error]:   in `example1.rb`:5:

	5: private_constant(*constants(false))

[error]: NameError: uninitialized constant YARD::Handlers::Ruby::PrivateConstantHandler::NamespaceMissingError
[error]: Stack trace:
	/Users/johnbackus/.rvm/gems/ruby-2.2.3/gems/yard-0.8.7.6/lib/yard/handlers/ruby/private_constant_handler.rb:33:in `rescue in privatize_constant'
	/Users/johnbackus/.rvm/gems/ruby-2.2.3/gems/yard-0.8.7.6/lib/yard/handlers/ruby/private_constant_handler.rb:25:in `privatize_constant'
	/Users/johnbackus/.rvm/gems/ruby-2.2.3/gems/yard-0.8.7.6/lib/yard/handlers/ruby/private_constant_handler.rb:11:in `block (2 levels) in <class:PrivateConstantHandler>'
	/Users/johnbackus/.rvm/gems/ruby-2.2.3/gems/yard-0.8.7.6/lib/yard/handlers/ruby/private_constant_handler.rb:8:in `each'
	/Users/johnbackus/.rvm/gems/ruby-2.2.3/gems/yard-0.8.7.6/lib/yard/handlers/ruby/private_constant_handler.rb:8:in `block in <class:PrivateConstantHandler>'
	/Users/johnbackus/.rvm/gems/ruby-2.2.3/gems/yard-0.8.7.6/lib/yard/handlers/processor.rb:114:in `block (2 levels) in process'

Files:           1
Modules:         0 (    0 undocumented)
Classes:         1 (    1 undocumented)
Constants:       0 (    0 undocumented)
Methods:         0 (    0 undocumented)
 0.00% documented
[...]

PR #974, opened 2016-06-20
Fix NameError in PrivateClassMethodHandler

SuperTux88:
Same error as in #902, but for `private_class_method`.

It happens when you write `private_class_method def self.some_private_class_method`.

```
[error]: NameError: uninitialized constant YARD::Handlers::Ruby::PrivateClassMethodHandler::NamespaceMissingError
```


lsegal:
Thanks!

PR #977, opened 2016-06-29
Allow wrapping extra file meta data in HTML comment

tf:
Meta data for extra files has to be placed at the top of the
file. Since each line has to be prefixed with a `#` those lines are
displayed as headers when viewing markdown files on GitHub. This
commit allows hiding the meta data section in rendered markdown on
GitHub by wrapping it in an HTML comment:

    

    Rest of document


lsegal:
Looks great, thanks @tf!

PR #978, opened 2016-07-09
sort files

boutil:
This should make the order of files (more) deterministic, and help with creating reproducible documentation.

### DIFF
--- a/lib/yard/cli/yardoc.rb
+++ b/lib/yard/cli/yardoc.rb
@@ -278,8 +278,8 @@ module YARD
         # Last minute modifications
         self.files = Parser::SourceParser::DEFAULT_PATH_GLOB if self.files.empty?
         self.files.delete_if {|x| x =~ /\A\s*\Z/ } # remove empty ones
-        readme = Dir.glob('README{,*[^~]}').first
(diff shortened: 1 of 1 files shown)


lsegal:
Can you add some tests for this?

lib/yard/cli/yardoc.rb
```diff
@@ -393,6 +393,7 @@ def print_list
       # @param [Array<String>] files the set of documentation files
       def add_extra_files(*files)
         files.map! {|f| f.include?("*") ? Dir.glob(f) : f }.flatten!
+        files.sort!
```
lsegal: This would change file order when files were listed explicitly, which is currently a supported feature. It's fairly common to want to manually sort extra files, for example, like putting your Getting Started Guide before more advanced documentation-- this is especially useful in the guide template, where document order is purposeful: https://github.com/lsegal/yard/blob/master/.yardopts_guide#L9

Sorting makes sense for a README, and even for file globbing, but not so much for the finalized list.

lsegal:
Marking this as closed since it changes expected behavior. If you want to take a look at the comments and update the patch I'd accept a new PR!

PR #981, opened 2016-07-15
Exclude frozen string comment from file docstring

xtian:
Ruby 2.3 added the `# frozen_string_literal: true` magic comment which enables immutable strings for a particular file.

This PR excludes those comments from file documentation in the same way that shebangs and encoding comments are currently excluded.

Edit: I'm confused why CI is failing as all specs pass for me locally. Any advice?

### DIFF
--- a/lib/yard/parser/ruby/ruby_parser.rb
+++ b/lib/yard/parser/ruby/ruby_parser.rb
@@ -5,6 +5,7 @@ module YARD
     module Ruby
       # Ruby 1.9 parser
       # @!attribute [r] encoding_line
+      # @!attribute [r] frozen_string_line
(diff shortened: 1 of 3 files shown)


lsegal:
Thanks for the PR! You probably want to put this test in the ruby/ruby_parser_spec.rb file so it doesn't test against legacy parsing.

lsegal:
FYI you can test in legacy mode by setting the `LEGACY` env var to any value (`LEGACY=1 rake`) to test this locally.

lsegal:
Thanks for the PR!

xtian:
@lsegal Thanks for the help! Cheers

PR #989, opened 2016-07-19
Add since tag (0.9.1) to RegistryResolver class

MSP-Greg:


MSP-Greg:
I've got a stupid GitHub question.  Where are the commits for --

`lib/yard/registry_resolver.rb`  and `lib/yard/code_objects/namespace_mapper.rb`?

Could info about new/removed modules/classes be added to CHANGELOG.md?

lsegal:
The changelog entry discusses the changes here: https://github.com/lsegal/yard/blame/master/CHANGELOG.md#L26-L28

The two classes don't actually change behavior so they were omitted. You're not expected to use either class directly (this is explicitly mentioned in the RegistryResolver docs), so mentioning them in the changelog would be counter-productive.

PR #1000, opened 2016-07-27
CSS Updates for UI - style.css and full_list.css

MSP-Greg:
### Description

Changes to the two css files, style.css and full_list.css.
#### Changes regardless of browser window width:
- List pane scrollbar does not extend into header.
- Momentum scrolling added for both main doc & list pane.
- List pane header adjusted to avoid incorrect search text input box placement in iOS.
- Enlarged the target area for 'Class List' tree open/close.  On a tablet, one has a fighting chance of opening a 'Class List' node without also loading the node.
#### Changes for browser window width less that 920 pixels in width:
- List pane show/hide control (a#full_list_link) is locked to the right window side, not the right document side.  Previously, for 'over-width' docs, it would require horizontal scrolling to make the control visible.
- List pane is limited in height to the window height.
- Changes the layout of constant code block, it will now wrap text.  Previously, many constants that were long lists of items would cause the doc page to   overflow the window width.  This corrects that, but may cause issues when a  constant is defined by lines of code wider than the screen. View the following in a narrow browser window:
  http://www.rubydoc.info/github/lsegal/yard/YARD/CodeObjects
#### General changes:
- Correctly ordered vendor prefixed css properties, of course, per convention.
#### Tested with:
- Current desktop versions of Chrome and Firefox
- iOS Safari and Chrome
### Completed Tasks
- [X] I have read the [Contributing Guide](https://github.com/lsegal/yard/blob/master/CONTRIBUTING.md).
- [ ] NOT APPLICABLE - I wrote tests and they pass (if code is attached to PR).

### DIFF
--- a/templates/default/fulldoc/html/css/style.css
+++ b/templates/default/fulldoc/html/css/style.css
@@ -1,4 +1,4 @@
-html {
+html { margin: 0; padding: 0;
   width: 100%;
   height: 100%;
 }
@@ -6,11 +6,13 @@ body {
   font-family: "Lucida Sans", "Lucida Grande", Verdana, Arial, sans-serif;
   font-size: 13px;
   width: 100%;
+  height: 100%;
   margin: 0;
   padding: 0;
-  display: flex;
   display: -webkit-flex;
   display: -ms-flexbox;
+  display: flex;
+  overflow: hidden;
 }
 
 #nav {
@@ -27,12 +29,12 @@ body {
   width: 20%;
   height: 100%;
   position: relative;
-  display: flex;
   display: -webkit-flex;
   display: -ms-flexbox;
-  flex-shrink: 0;
+  display: flex;
   -webkit-flex-shrink: 0;
   -ms-flex: 1 0;
+  flex-shrink: 0;
 }
 #resizer {
   position: absolute;
@@ -44,14 +46,17 @@ body {
   z-index: 9999;
 }
 #main {
-  flex: 5 1;
   -webkit-flex: 5 1;
   -ms-flex: 5 1;
+  flex: 5 1;
   outline: none;
   position: relative;
   background: #fff;
   padding: 1.2em;
   padding-top: 0.2em;
+  box-sizing: border-box;
+  overflow: auto;
+  -webkit-overflow-scrolling: touch;  
 }
 
 @media (max-width: 920px) {
@@ -59,8 +64,6 @@ body {
 }
 
 @media (min-width: 920px) {
-  body { height: 100%; overflow: hidden; }
-  #main { height: 100%; overflow: auto; }
   #search { display: none; }
   #search_frame { display: none; }
 }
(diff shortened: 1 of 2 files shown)


MSP-Greg:
Forgot to mention, demo of YARD using the PR css at
https://msp-greg.github.io/yard-no-t2/

MSP-Greg:
I noticed that you're doing some work on the css & js.

Should I close this or would you like to leave it around for a bit?

BTW, re yard-t2 css, html, & js, much of it I need to modify to bring into YARD due to the structural changes I've made.

One example, re the Class List
* Created a larger 'node open/close' target for nested lists, hence different structure.  I tried with the existing YARD structure, but just couldn't do it.  The target now extends all the way to the left node margin, regardless of level.
* Moved the indenting to css.
* Simplified the class list used on the nodes.
* Since loading is now via XMLHttpRequest, removing the iFrame changed the structure, css, & js.

As a general change, all the graphics are svg.

Finally, if you'd like any help, I'd be happy to check changes you make on the browser side.

templates/default/fulldoc/html/css/style.css
```diff
@@ -1,16 +1,18 @@
-html {
+html { margin: 0; padding: 0;
```
lsegal: This looks like a formatting issue.

lsegal:
I'm having trouble giving this a proper review because there are a number of changes and it's fairly difficult to figure out which changes are meant for which bullet points.

It would be easier to merge this as individual changes rather than wholesale. Could you do me a favor and split this PR up into separate PRs for smaller separable changes (scrolling, controls, height issues, etc) so I can see how these rules are being applied?

Thanks!

MSP-Greg:
Yes, I can do so.  I'd appreciate if you could check it 'as is' on (if possible) a desktop & tablet.

I'd like to know -

* Does it appear to work
* Is there anything you don't want changed

templates/default/fulldoc/html/css/style.css
MSP-Greg: Not sure what you mean.  You want the properties split across two lines?

lsegal: Yes. The rest of this rule is formatted as a declaration on a separate line. This change seems to break that convention.

lsegal:
It does appear to work on a desktop (I don't own an iOS device), though I noticed minor issues that can be discussed in respective PRs. As for anything not being changed-- it's hard to know as a complete set. That said, it does look like the last item:

> Changes the layout of constant code block, it will now wrap text. 

Will probably not be accepted in its current form. The change seems to remove code pre-formatting, which means anything with newlines will become garbled. Consider `ENCODING_BYTE_ORDER_MARKS` in the two URLs:

Original: http://www.rubydoc.info/github/lsegal/yard/YARD/Parser/SourceParser#ENCODING_BYTE_ORDER_MARKS-constant

vs.

Proposal: https://msp-greg.github.io/yard-no-t2/YARD/Parser/SourceParser.html#ENCODING_BYTE_ORDER_MARKS-constant

If you can make pre-formatting work with wrapping, feel free to open that separately.

MSP-Greg:
Apology, somehow I missed this message.

Re the constant & cvar source/value blocks, I did change that to show a horizontal scrollbar, rather than trying to force a wrap.

I will get back to this, but I'll probably look at that odd parser issue first.

Issue #1003, opened 2016-07-29
YARD::CodeObjects::BUILTIN_EXCEPTIONS - missing exception classes


MSP-Greg:
The following root exception classes are missing from BUILTIN_EXCEPTIONS 

```ruby
ClosedQueueError
EncodingError
FiberError
KeyError
StopIteration
UncaughtThrowError
```

Replacement code (sorted)

```ruby
    BUILTIN_EXCEPTIONS = ["ArgumentError", "ClosedQueueError", "EncodingError",
      "EOFError", "Exception", "FiberError", "FloatDomainError", "IndexError",
      "Interrupt", "IOError", "KeyError", "LoadError", "LocalJumpError",
      "NameError", "NoMemoryError", "NoMethodError", "NotImplementedError",
      "RangeError", "RegexpError", "RuntimeError", "ScriptError", "SecurityError",
      "SignalException", "StandardError", "StopIteration", "SyntaxError",
      "SystemCallError", "SystemExit", "SystemStackError", "ThreadError",
      "TypeError", "UncaughtThrowError", "ZeroDivisionError"]
```

* OS: Windows
* Ruby version 2.3.2 & 2.4.0
* YARD version 0.9.5
* Relevant software dependency/versions:
  * NA

I have read the [Contributing Guide][contrib].

[contrib]: https://github.com/lsegal/yard/blob/master/CONTRIBUTING.md

lsegal:
Can you pull request this?

MSP-Greg:
Yes to both this and #1004.  Soon, but not immediately.

lsegal:
Closed by #1005

PR #1005, opened 2016-07-29
Update to CodeObjects 'BUILTIN' constants

MSP-Greg:
# Description

Per #1003. Additionally, I decided to update all three,  `BUILTIN_EXCEPTIONS`, `BUILTIN_CLASSES`, and `BUILTIN_MODULES`.

* Ordered items
* Switched to %w
* Ruby 2.3.2
* Left `MatchingData` in `BUILTIN_CLASSES`
* `BUILTIN_CLASSES` still doesn't contain `ARGF` & `ENV` , both fail RSpec
* Kept approx 80+ char margin
* Blank space between ok?

# Completed Tasks

* [ ] Git commits have been cleaned up (squash WIP / revert commits).
* [X] I wrote tests and they pass (if code is attached to PR).

### DIFF
--- a/spec/code_objects/constants_spec.rb
+++ b/spec/code_objects/constants_spec.rb
@@ -50,19 +50,27 @@ describe YARD::CodeObjects do
 
   describe :BUILTIN_EXCEPTIONS do
     it "includes all base exceptions" do
+      bad_names = []
       YARD::CodeObjects::BUILTIN_EXCEPTIONS.each do |name|
-        expect(eval(name)).to be <= Exception
+        begin
+          bad_names << name unless eval(name) <= Exception
+        rescue NameError
+        end
       end
+      expect(bad_names).to be_empty
     end
   end
 
   describe :BUILTIN_CLASSES do
     it "includes all base classes" do
+      bad_names = []
       YARD::CodeObjects::BUILTIN_CLASSES.each do |name|
-        next if name == "MatchingData" && !defined?(::MatchingData)
-        next if name == "Continuation"
-        expect(eval(name)).to be_instance_of(Class)
+        begin
+          bad_names << name unless eval(name).is_a?(Class)
+        rescue NameError
+        end
       end
+      expect(bad_names).to be_empty
     end
 
     it "includes all exceptions" do
--- a/lib/yard/code_objects/base.rb
+++ b/lib/yard/code_objects/base.rb
@@ -63,20 +63,29 @@ module YARD
     METHODMATCH = /(?:(?:#{NAMESPACEMATCH}|[a-z]\w*)\s*(?:#{CSEPQ}|#{NSEPQ})\s*)?#{METHODNAMEMATCH}/
 
     # All builtin Ruby exception classes for inheritance tree.
-    BUILTIN_EXCEPTIONS = ["SecurityError", "Exception", "NoMethodError", "FloatDomainError",
-      "IOError", "TypeError", "NotImplementedError", "SystemExit", "Interrupt", "SyntaxError",
-      "RangeError", "NoMemoryError", "ArgumentError", "ThreadError", "EOFError", "RuntimeError",
-      "ZeroDivisionError", "StandardError", "LoadError", "NameError", "LocalJumpError", "SystemCallError",
-      "SignalException", "ScriptError", "SystemStackError", "RegexpError", "IndexError"]
+    BUILTIN_EXCEPTIONS = ["ArgumentError", "ClosedQueueError", "EncodingError",
+      "EOFError", "Exception", "FiberError", "FloatDomainError", "IndexError",
+      "Interrupt", "IOError", "KeyError", "LoadError", "LocalJumpError",
+      "NameError", "NoMemoryError", "NoMethodError", "NotImplementedError",
+      "RangeError", "RegexpError", "RuntimeError", "ScriptError", "SecurityError",
+      "SignalException", "StandardError", "StopIteration", "SyntaxError",
+      "SystemCallError", "SystemExit", "SystemStackError", "ThreadError",
+      "TypeError", "UncaughtThrowError", "ZeroDivisionError"]
+
     # All builtin Ruby classes for inheritance tree.
     # @note MatchingData is a 1.8.x legacy class
-    BUILTIN_CLASSES = ["TrueClass", "Array", "Dir", "Struct", "UnboundMethod", "Object", "Fixnum", "Float",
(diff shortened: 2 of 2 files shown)


lsegal:
Tests don't seem to be passing on earlier Ruby versions (specs probably need updating), and I see some comment formatting errors. I think we should just start with the added constants to deal with #1003 directly and then open a refactoring PR afterwards if you want to change the coding style so as to not conflate these two changes.

MSP-Greg:
Sorry about that comment, my damn editor does that (add a 'spacer' line, and it indents the next).  Missed it.

Re the tests, it seems the only error is related to ClosedQueueError, which is new in 2.3 and not defined in 2.2 or earlier (I've got 2.0, 2.2, 2.3, and 2.4 doc'd).

The two tests expect all BUILTIN members to be defined in the running version of Ruby.
```ruby
expect(eval(name)).to be_instance_of(Class)
expect(eval(name)).to be <= Exception
```

* Should I leave that in?
* %w or string array?

Sorry, I didn't look at the tests, I just saw they passed with Ruby 2.3.2.

Also, I've wondered about having multiple versions of Ruby installed, but haven't yet done so.  You use Windows?  If so, reasonably easy to set up that way?

lsegal:
We need to fix the tests to be lenient regarding missing constants on older versions or the tests will not pass.

Use a string array for now, if you want to refactor, I can look at that separately.

As for running tests, you should be able to see Travis output on your own fork if you set it up. If you want to run locally, I'd recommend [installing Docker](https://docs.docker.com/engine/installation/windows/) and just running:

```sh
> docker run -v %CD%:/yard -w /yard ruby:2.2 /bin/sh -c "bundle && bundle exec rake"
```

MSP-Greg:
FYI, I checked for what's added from the previous code, re RSpec, and came up with the following (ClosedQueueError just happens to be on both lists and is the 1st addition alphabetically) --

new Exception Classes
```ruby
ClosedQueueError
EncodingError
FiberError
KeyError
StopIteration
UncaughtThrowError
```

new Classes
```ruby
Complex
ConditionVariable
Encoding
Enumerator
Fiber
Queue
Random
RubyVM
SizedQueue
TracePoint
```

MSP-Greg:
@lsegal  I'm having a bad git day (stack patch).  IMHO, git docs (git-scm, etc) are very good when they address working  with one's own repro and a remote origin, but they get sketchy when dealing with forks, etc, especially how to have multiple PR's open between a local fork, a remote origin, and the original repo.  I should have never used the master branch for these commits/PR...

Anyway,

> We need to fix the tests to be lenient regarding missing constants on older versions or the tests will not pass.

I'm looking over some of the Ruby 2.4 docs, I suspect we need the tests to work both ways, as there may be Classes/Exceptions/Modules that might be deprecated or removed in future versions.  Also, the tests currently fail on first failure, it would be nice if they loaded an array with all failures and then tested  whether the array was empty, hence, showing all failures, not just the first.

lsegal:
> I'm looking over some of the Ruby 2.4 docs, I suspect we need the tests to work both ways, as there may be Classes/Exceptions/Modules that might be deprecated or removed in future versions. 

I **strongly** doubt any class in Ruby will ever go away, regardless of how "deprecated" it is (technically none of those are even deprecated yet). It's fairly safe to keep any existing constant in the tests. Given that it's only a testing functionality (YARD doesn't eval these during runtime), any failures would be caught as we move Travis to new Ruby versions. A RUBY_VERSION check for new constants can be used to exclude the new constants, and same for old ones:

```ruby
it 'matches these constants' do
  ...
end if RUBY_VERSION >= '2.4'
```

> Also, the tests currently fail on first failure, it would be nice if they loaded an array with all failures and then tested whether the array was empty, hence, showing all failures, not just the first.

I'd accept this change. Moving the class loop outside of the it() declaration (one it() block per class) would fix this.

MSP-Greg:
I'm currently working on this.  So we'll list constants for 2.3, and test for earlier and later.

Reason for later - `Mutex` exists in 2.2, changed to `Thread::Mutex` in 2.3.  Yes, I'm (not) having fun.  Are you sure you only want to have 2.2.x and 2.3.x in travis.yml?  2.1? 2.0?

lsegal:
Running tests on Travis against 2.0 and 2.1 is difficult now with dependency changes (like Rack, etc). Given the EOL for both versions, I'm okay with stopping at a "best effort" to ensure compatibility, and if regressions occur, we can accept patches to fix them. The burden on maintaining CI against those old versions is too high.

MSP-Greg:
@lsegal Eight days ago I was looking at this, but I moved to other issues.  Can we revisit this?

I've never closed a PR, if I do so, does that flush everything out?  Again, apology for PR'ing on the main branch.

I'd like to PR the stack `freeze_tree` code, that test string flip may still be around, I've got a patch to the frozen strings 'directive', etc.

spec/code_objects/constants_spec.rb
```diff
@@ -50,19 +50,30 @@
 
   describe :BUILTIN_EXCEPTIONS do
     it "includes all base exceptions" do
+      bad_names = []
       YARD::CodeObjects::BUILTIN_EXCEPTIONS.each do |name|
-        expect(eval(name)).to be <= Exception
+         next if name == 'ClosedQueueError' && RUBY_VERSION < '2.3'
```
lsegal: Instead of manually mapping out constants (this will be a headache to maintain), why not just ignore any constant that does not exist? A NameError rescue should suffice.

spec/code_objects/constants_spec.rb
```diff
@@ -50,19 +50,30 @@
...
+         bad_names << name unless eval(name) <= Exception
       end
+      expect(bad_names).to be_empty
     end
   end
 
   describe :BUILTIN_CLASSES do
     it "includes all base classes" do
+      if RUBY_VERSION < '2.3'
```
lsegal: Ditto for manual checks. It shouldn't matter if the list includes more than the base classes (for that version).

lsegal:
@MSP-Greg I added some comments, but this also needs the commit history cleaned up. You can [rebase and force push](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) your branch after you clean it up so that you don't have to close the PR and open a new one. Ideally this should only require 1 single commit.

MSP-Greg:
Thank you for the Git tips.  I generally don't forget once I've done something.

lsegal:
Looks good, thanks!

PR #1008, opened 2016-08-09
Server Process.fork error Windows

MSP-Greg:
After pulling changes tonite, tests failed on Windows.

Crash on `Process.fork`.  See the following, and note

'If fork is not usable, Process.respond_to?(:fork) returns false.'  Hence, added that test to the code.

https://msp-greg.github.io/Ruby-2.3/Core/Process.html#fork-class_method

# Completed Tasks

* [X] Tested locally against Ruby 2.2.4 & 2.3.2.

### DIFF
--- a/lib/yard/server/commands/library_command.rb
+++ b/lib/yard/server/commands/library_command.rb
@@ -29,11 +29,13 @@ module YARD
       #
       # @abstract
       class LibraryCommand < Base
-        begin
-          Process.fork { exit 0 }
-          CAN_FORK = true
-        rescue
-          CAN_FORK = false
+        CAN_FORK = false
+        if Process.respond_to?(:fork)
+          begin
+            Process.fork { exit 0 }
+            CAN_FORK = true
+          rescue
+          end
         end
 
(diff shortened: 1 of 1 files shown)


lsegal:
Can you explain how this failed (stack trace, etc)? I tested the original code on a Windows 10 machine and it seemed to be working fine; the rescue should catch the exception if one is raised. The `respond_to?` is not reliable across Ruby implementations (that doc note only applies to MRI, not JRuby, for example), which is why it was omitted.

I'd be curious to see how your tests were failing.

lsegal:
Ah, I was able to reproduce. The rescue should have been rescuing `Exception` instead of a blank rescue call (technically a StandardError might be better). This makes it so the `respond_to?` is not needed (and will still handle an implementation that responds to the method but still fails).

MSP-Greg:
Your commit passes local tests on 2.2.4 & 2.3.2.

Thanks, I saw the mention in the docs and forgot about rescue, and how it works 'a bit different' in Ruby...

Issue #1002, opened 2016-07-29
Advice needed: "stack level too deep"


zverok:
Hey.

While trying to document really large codebase, I stuck with
```
yard-0.9.5/lib/yard/parser/ruby/ast_node.rb:198:in `children': stack level too deep (SystemStackError)
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/gems/yard-0.9.5/lib/yard/parser/ruby/ruby_parser.rb:650:in `freeze_tree'
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/gems/yard-0.9.5/lib/yard/parser/ruby/ruby_parser.rb:652:in `block in freeze_tree'
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/gems/yard-0.9.5/lib/yard/parser/ruby/ruby_parser.rb:650:in `each'
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/gems/yard-0.9.5/lib/yard/parser/ruby/ruby_parser.rb:650:in `freeze_tree'
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/gems/yard-0.9.5/lib/yard/parser/ruby/ruby_parser.rb:652:in `block in freeze_tree'
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/gems/yard-0.9.5/lib/yard/parser/ruby/ruby_parser.rb:650:in `each'
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/gems/yard-0.9.5/lib/yard/parser/ruby/ruby_parser.rb:650:in `freeze_tree'
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/gems/yard-0.9.5/lib/yard/parser/ruby/ruby_parser.rb:652:in `block in freeze_tree'
	 ... 10163 levels...
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/bin/yardoc:23:in `load'
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/bin/yardoc:23:in `<main>'
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/bin/ruby_executable_hooks:15:in `eval'
	from /media/storage/home/zverok/.rvm/gems/ruby-2.3.1/bin/ruby_executable_hooks:15:in `<main>'
```

[...]

PR #1009, opened 2016-08-09
Issue #1002 - 'stack level too deep' - freeze_tree

MSP-Greg:
Per discussion in issue #1002 - 'stack level too deep', change removes re-entrant code.

# Completed Tasks

* [X] Tested locally with Ruby 2.2.4 & 2.3.2.

### DIFF
--- a/lib/yard/parser/ruby/ruby_parser.rb
+++ b/lib/yard/parser/ruby/ruby_parser.rb
@@ -656,10 +656,13 @@ module YARD
         end
 
         def freeze_tree(node = nil)
-          node ||= root
-          node.children.each do |child|
-            child.parent = node
-            freeze_tree(child)
+          nodes = [node || root]
+          while !nodes.empty?
+            p_node = nodes.shift
+            p_node.children.each do |child|
+              child.parent = p_node
+              nodes << child
+            end
           end
         end
       end if defined?(::Ripper)


MSP-Greg:
I was testing this (which previously passed) when the Process.fork issue came up, and my 2.2.4 install was complaining about the (recent) changes to RSpec, Rake, etc.  Got it fixed...

lib/yard/parser/ruby/ruby_parser.rb
```diff
@@ -656,10 +656,12 @@ def add_comment(line, node = nil, before_node = nil, into = false)
...
 
         def freeze_tree(node = nil)
-          node ||= root
-          node.children.each do |child|
-            child.parent = node
-            freeze_tree(child)
+          nodes = []
+          nodes.push(node || root)
+          while !nodes.empty? do
+            p_node = nodes.shift
```
lsegal: Does it matter what the freeze order is? Calling `nodes.pop` should be much more efficient.

MSP-Greg:
> Does it matter what the freeze order is?

I just replaced the re-entrant code with something that behaved exactly the same way.  Aren't you supposed to know that answer?  Ok, just kidding...

So, shift mirrors the original code, iow, a parent has the operation (setting all its children's parent prop to itself) done before it's done to the children.  Using pop, siblings will still be on the 'left', while their children are pulled off of the right.

If you're sure that isn't a problem now or in the future, we (I?) can change it.  Either way, still has to be faster than re-entrant code...

lsegal:
> Either way, still has to be faster than re-entrant code...

The iterative (correct term for non-recursive) implementation is actually slower (even with pop), though using pop makes the overhead much less significant:

```
$ ruby benchmarks/freeze_tree.rb 50
Rehearsal -----------------------------------------------
recursive     4.910000   0.010000   4.920000 (  4.946792)
iter-shift    5.780000   0.130000   5.910000 (  6.030129)
iter-pop      5.090000   0.000000   5.090000 (  5.140247)
------------------------------------- total: 15.920000sec

                  user     system      total        real
recursive     4.560000   0.000000   4.560000 (  4.708833)
iter-shift    5.780000   0.000000   5.780000 (  5.878878)
iter-pop      4.850000   0.010000   4.860000 (  4.978221)
```

Listing of benchmarks/freeze_tree.rb:

```ruby
require 'benchmark'
require_relative '../lib/yard'
require_relative '../lib/yard/parser/ruby/ruby_parser'

root = File.join(File.dirname(__FILE__), '..')
file = File.expand_path('lib/yard/parser/ruby/legacy/ruby_lex.rb', root)
src  = File.read(file) 

TIMES = (ARGV[0] || 10_000).to_i

class ParseRecursive < YARD::Parser::Ruby::RipperParser
  def freeze_tree(node = nil)
    node ||= root
    node.children.each do |child|
      child.parent = node
      freeze_tree(child)
    end
  end
end

class ParseIterShift < YARD::Parser::Ruby::RipperParser
  def freeze_tree(node = nil)
    nodes = []
    nodes.push(node || root)
    while !nodes.empty? do
      p_node = nodes.shift
      p_node.children.each { |child| child.parent = p_node }
      nodes += p_node.children
    end
  end
end

class ParseIterPop < YARD::Parser::Ruby::RipperParser
  def freeze_tree(node = nil)
    nodes = []
    nodes.push(node || root)
    while !nodes.empty? do
      p_node = nodes.pop
      p_node.children.each { |child| child.parent = p_node }
      nodes += p_node.children
    end
  end
end

Benchmark.bmbm do |b|
  b.report("recursive") { TIMES.times { ParseRecursive.new(src, file).parse } }
  b.report("iter-shift") { TIMES.times { ParseIterShift.new(src, file).parse } }
  b.report("iter-pop") { TIMES.times { ParseIterPop.new(src, file).parse } }
end
```

lib/yard/parser/ruby/ruby_parser.rb
```diff
@@ -656,10 +656,12 @@ def add_comment(line, node = nil, before_node = nil, into = false)
         end
 
         def freeze_tree(node = nil)
-          node ||= root
-          node.children.each do |child|
-            child.parent = node
-            freeze_tree(child)
+          nodes = []
+          nodes.push(node || root)
```
lsegal: Instead of the separate push call, this should just be `nodes = [node || root]`.

lib/yard/parser/ruby/ruby_parser.rb
```diff
@@ -656,10 +656,12 @@ def add_comment(line, node = nil, before_node = nil, into = false)
...
+          while !nodes.empty? do
```
lsegal: YARD formatting style does not use a `do` on while loop blocks by convention.

MSP-Greg:
Try the following, both were equal or faster than recursive on CodeObjects::Base
```
class ParseIterShift < YARD::Parser::Ruby::RipperParser
  def freeze_tree(node = nil)
    nodes = [node || root]
    while !nodes.empty? {
      p_node = nodes.shift
      p_node.children.each { |child|
        child.parent = p_node
        nodes << child
      }
    }
  end
end

class ParseIterPop < YARD::Parser::Ruby::RipperParser
  def freeze_tree(node = nil)
    nodes = [node || root]
    while !nodes.empty? {
      p_node = nodes.pop
      p_node.children.each { |child|
        child.parent = p_node
        nodes << child
      }
    }
  end
end
```

Your suggestion of `nodes = [node || root]` helped, but changing from
`nodes += p_node.children`
to 
`nodes << child`
was what made them faster.  Not expected.  Interested to see your results.

MSP-Greg:
Sorry, edited code after running tests, correct code on GitHub, not email.   I tested a bit more more and I found no significant difference between the three.

MSP-Greg:
PR / commit updated

lib/yard/parser/ruby/ruby_parser.rb
```diff
@@ -656,10 +656,13 @@ def add_comment(line, node = nil, before_node = nil, into = false)
...
 
         def freeze_tree(node = nil)
-          node ||= root
-          node.children.each do |child|
-            child.parent = node
-            freeze_tree(child)
+          nodes = [node || root]
+          while !nodes.empty?
+            p_node = nodes.shift
+            p_node.children.each { |child|
```
lsegal: Update to `do .. end` for multiline blocks and this is good to merge. Thanks!

lsegal:
Thanks for your work on this!

PR #1013, opened 2016-08-11
Parser percent arrays fix

MSP-Greg:
# Description

Fixes percent array parsing with `\n ]` for closing delimiter
# Completed Tasks
- [X] I have read the [Contributing Guide](https://github.com/lsegal/yard/blob/master/CONTRIBUTING.md).
- [X] I wrote tests and they pass (if code is attached to PR).

### DIFF
--- a/spec/parser/ruby/ruby_parser_spec.rb
+++ b/spec/parser/ruby/ruby_parser_spec.rb
@@ -283,6 +283,58 @@ eof
       end
     end
 
+    it "\n     Parses several formats of percent arrays into source strings\n" \
+         "     in constant declarations.  Loops thru all combinations of\n" \
+         "     () {} [] <> || delimiters, %i %I %w %W array types, and five\n" \
+         "     types of element/delimiter whitespace." do
+      delimiter_h = {
+        '(' => ')',
+        '{' => '}',
+        '[' => ']',
+        '<' => '>',
+        '|' => '|'
+      }
+
+      type_h = {
+        'i' => :qsymbols_literal,
+        'I' => :symbols_literal,
+        'w' => :qwords_literal,
+        'W' => :words_literal
+      }
+
+      str_a = [
+        'TEST = %{type}[A B C]',
+        'TEST = %{type}[  A  B  C  ]',
+        'TEST = %{type}[ \nA \nB \nC \n]',
+        'TEST = %{type}[\n\nAD\n\nB\n\nC\n\n]',
+        'TEST = %{type}[\n A\n B\n C\n ]'
+      ]
+
+      failed_tests = []
+
+      delimiter_h.each do |o, c|
+        type_h.each do |id, sym|
+          str_a.each do |str|
+            str_sub = str.sub(/\{type\}/, id).sub(/\[/, o).sub(/\]/, c)
+
+            s = stmt(str_sub.gsub(/\\n/, "\n"))
+
+            node = s.jump(sym)
+            p_source = node.source.gsub(/\n/, '\\n')
+            if p_source != str_sub[7..-1]
+              failed_tests << "#{str_sub.ljust(31)} ->  #{p_source}"
+            end
+          end
+        end
+        failed_tests << "\n" unless failed_tests.empty?
+      end
+      unless failed_tests.empty?
+        msg = "Failures:\n#{'-------- INPUT --------'.ljust(36)}" \
+          "---- OUTPUT -----\n#{failed_tests.join("\n")}"
+      end
+      expect(failed_tests).to be_empty, msg
+    end
+
     it "parses %w() array in constant declaration" do
       s = stmt(<<-eof)
         class Foo
--- a/lib/yard/parser/ruby/ruby_parser.rb
+++ b/lib/yard/parser/ruby/ruby_parser.rb
@@ -49,6 +49,7 @@ module YARD
           @frozen_string_line = nil
           @file_encoding = nil
           @newline = true
+          @percent_ary = nil
         end
 
         def parse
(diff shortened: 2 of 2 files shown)


spec/parser/ruby/ruby_parser_spec.rb
```diff
@@ -283,6 +283,54 @@ def hello; end
       end
     end
 
+    it "correcly parses several formats of percent arrays in constant declarations" do
```
lsegal: You can rewrite this single test as a series of `it()` blocks (move the `it` call into the `str_a.each` block). That way you don't have to do your own failure output management at the bottom-- rspec can handle this automatically.

MSP-Greg:
Thanks for the explanation of some options.

Personally, I would prefer to have the error message formatted as is, since, by listing on consecutive lines, it's very clear where the parsing error is.  For instance, what if in the future Ripper separates the 'left side delimiter' from the %i, %I, %w , or %W characters?  The failure output would make it very clear.

I'll admit to having a bias towards tabular data.  I think one comment thread I alluded to that regarding log messages...

Hence, what would you like me to do?

lsegal:
I would recommend making the `it()` change. The extra granularity on the tests allows for better understanding of which tests failed, and allows us to run individual tests rather than all of them together (very useful when debugging a specific parsing issue with logging). If you really want the output text, you can always add the `"#{str_sub.ljust(31)} ->  #{p_source}"` data to the msg argument of the individual `expect()` calls. Best of both worlds.

spec/parser/ruby/ruby_parser_spec.rb
```diff
@@ -283,6 +283,54 @@ def hello; end
...
+      delimiter_a = [
+        ['(', ')'],
+        ['{', '}'],
+        ['[', ']'],
+        ['<', '>'],
+        ['|', '|']
+      ]
+
+      type_a = [
```
lsegal: This may as well be a hash for readability (since order doesn't matter for 1.8 anyway)

MSP-Greg:
First, I changed both the arrays of two element arrays to hashes. then I changed the block structure, then commented out the one line in ruby_parser.  Ran RSpec, and honestly, I thought it was a mess.

We've got a 5 (delimiters) x 4 (%i, %I, etc) x 5 (whitespace) set of tests, and chances are, one item in one dimension/vector may fail in the future.  So we'll probably have at least 20 or 25 failures.  Then again, could be more, or could be less.  Without the tabular format, it is almost impossible to collate the info into anything meaningful.

Hence, I left it as is, but I did update the description to make clear all the tests it performs.  We'll hope it 
doesn't fail for a while.

Tested, robo'd, commit rebased and pushed...

lsegal:
> We've got a 5 (delimiters) x 4 (%i, %I, etc) x 5 (whitespace) set of tests, and chances are, one item in one dimension/vector may fail in the future.

Why would these tests fail in the future? If we're expecting them to fail at some point, then they should be re-written to be less brittle. If you're suggesting that a significant change in the parser may make all of them fail, I have no problem with this.

Ultimately this is not really acceptable in the current form. `it` descriptions by convention should not contain paragraph-long descriptions or newlines. Such a long it description really proves that these are multiple independent tests. I would accept this patch if the test is split up as recommended.

spec/parser/ruby/ruby_parser_spec.rb
```diff
@@ -283,6 +283,58 @@ def hello; end
...
+
+      type_h = {
+        'i' => :qsymbols_literal,
+        'I' => :symbols_literal,
+        'w' => :qwords_literal,
+        'W' => :words_literal
+      }
+
+      str_a = [
+        'TEST = %{type}[A B C]',
```
lsegal: Testing each delimiter type seems unnecessary. There is no separate code path for each delimiter in YARD's codebase, so this is a detail of the underlying library and doesn't need testing. In other words, testing for a single delimiter should be sufficient-- if a separate delimiter fails, that would not be a bug with this patch (there is no logic for delimiters in the parser).

spec/parser/ruby/ruby_parser_spec.rb
```diff
@@ -283,6 +283,58 @@ def hello; end
       end
     end
 
+    it "\n     Parses several formats of percent arrays into source strings\n" \
+         "     in constant declarations.  Loops thru all combinations of\n" \
+         "     () {} [] <> || delimiters, %i %I %w %W array types, and five\n" \
+         "     types of element/delimiter whitespace." do
+      delimiter_h = {
```
lsegal: Testing each delimiter is unnecessary. There is no codepath for delimiters in any of the parser code. Testing with a single delimiter should be sufficient-- if `%w()` worked and `%w{}` failed, that would not be due to a bug in YARD's codebase.

MSP-Greg:
@lsegal I'll revise this.  Been busy with some other things.

> If you're suggesting that a significant change in the parser may make all of them fail, I have no problem with this.

Certainly not all of them.  I'm suggesting, having worked on the percent array bug and the unary parameter bug, that Ripper is designed more for one-way parsing, rather than returning info for source code 'dice and chop'.  Hence, somewhat brittle for its application in YARD.

I'll revise my work files and create a repo for them, and keep the full tests in there.  If in the future, changes in Ripper result in some bugs creeping back in, at least the full tests will still be available.  Of course, they may not be helpful...

lsegal:
Merged as c9c9c0aca8bd9d4dbc39775855c4e129e70b108f with tests fixes at c27a05b8ea73fbb9aebb67fe4f238a2203cafc05

MSP-Greg:
@lsegal Sorry and thanks, been busy with Ruby & Windows...

Since you might be rev'ing YARD, when using Ruby 2.4/2.5, there's a few places  in `lib/yard/parser/ruby/legacy/ruby_lex.rb` where TRUE / FALSE generates depreciation warning.

Lines 643, 645, and 1167, 1168.  That's the only issues that I've seen running under 2.5.0...

lsegal:
Patches would be accepted for that, but the ruby_lex is 1.8 specific legacy code so I probably won't get around to this without a PR.

MSP-Greg:
I'll throw one together, not tonite.  The warnings happen on load/require, and I recall there was somewhere where a method of the class was used.  I think you even had a 'todo' type comment around the code (or the call).  Wasn't a tag though.

I may be confusing it with the legacy parser.  Maybe I'll try to find that code.

PR #1021, opened 2016-08-30
Add some ruby-lang builtin conventions

kachick:
# Description

Some ruby methods are expected to be overridden as `inspect`.
I'll be happy if I can reduce `@return [String]` from them as `def predicate?`.

# Completed Tasks

* [x] I wrote tests and they pass (if code is attached to PR).

---

But when this PR has any style difference, please tell me 🙇 
(I don't have confident around `/^$/` vs `/\A\z/` and `Hash syntax`...)

### DIFF
--- a/lib/yard/handlers/ruby/method_handler.rb
+++ b/lib/yard/handlers/ruby/method_handler.rb
@@ -1,6 +1,20 @@
 # frozen_string_literal: true
 # Handles a method definition
 class YARD::Handlers::Ruby::MethodHandler < YARD::Handlers::Ruby::Base
+  TYPE_CONVENTIONS = {
+    /\?\z/                                    => 'Boolean',
+    /\A(?:to_s|inspect|marshal_dump|_dump)\z/ => 'String',
+    /\A(?:to_int|to_i|hash)\z/                => 'Integer',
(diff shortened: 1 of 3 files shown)


lsegal:
Thanks for this PR, but this code is not likely to be accepted. The existing checks do not modify semantic behavior for the method's documentation.

This patch will force YARD to make semantic decisions about the documentation under a presumption that the author is following Ruby conventions. Although I'd love to presume that all Ruby authors follow all Ruby conventions, you cannot guarantee that common method names follow the same contracts. This is likely going to lead to many false positives, where YARD tells the user one thing, but the code does something else, which is even less helpful to users. I'd rather authors be explicit about the documentation they write.

This could be a useful plugin, though.

lsegal:
I should also add that another way to solve this would be an explicit opt-in via directive, like `@!inherit`, which would inherit a superclass's contracts-- alternatively just piggyback off of the `(see X)` syntax to do something like `(see super)`. YARD could then check superclasses, and a reliable solution could be to pull in the core method contracts for Ruby's core classes and use those as a base case.

kachick:
Thanks for your clarify!

> plugin
>@!inherit

Ah, sounds useful too 👍 
I'll try.

PR #1028, opened 2016-10-13
Replaced "backport" of `Gem.source_index` / `Gem::SourceIndex` with `Gem::Specification`

benlangfeld:
# Description

Because the former doesn't correctly load gems installed via `:path` or `:git` options in a Gemfile, meaning serving documentation for such gems via `yard server -G` throws silent errors (and spins infinitely in the UI).

The "backport" was introduced in https://github.com/lsegal/yard/commit/75e5aea51f149694269118ccd3ef6db09e7efc76 because RubyGems 1.8 removed SourceIndex and replaced it with Gem::Specification, but @lsegal didn't want to require YARD users to upgrade to Rubygems > 1.8. That was 5 years ago, and we should be able to move on now.

# Completed Tasks


benlangfeld:
The spec failure here was unrelated, fixed by https://github.com/bbatsov/rubocop/commit/3bcea86aec03114838492180713f03d45b2d42bd, and just needs an update to the installed gems on Travis. I figure they run a cache, so maybe we just have to wait.

lsegal:
This will technically be a breaking change. We could probably just toggle this based on the installed system RubyGems version.

benlangfeld:
> This will technically be a breaking change.

Sure, but only for those using ancient RubyGems versions, of which you can guess my opinion. Point taken though.

> We could probably just toggle this based on the installed system RubyGems version.

What version would you like to cut off at?

lsegal:
> What version would you like to cut off at?

Well instead of cutting off, you can just do feature detection:

```ruby
module YARD
  module GemIndex
    module_function
    def find_all_by_name(*args)
      if defined? Gem::Specification
        Gem::Specification.find_all_by_name(*args)
      else
        Gem.source_index.find_name(*args)
      end
    end
  end
end
```

Then just use `YARD::GemIndex.find_all_by_name` instead so there's less churn if we ever remove support for older RubyGems. And the `backports/*` would not need to be removed or touched.

benlangfeld:
> Well instead of cutting off, you can just do feature detection

Of course I can. Sorry, long day. Will prep this tomorrow.

benlangfeld:
Updated as requested @lsegal. Anything else you'd like changing?

lsegal:
If we get the changelog formatting right I can merge. Thanks for doing this!

benlangfeld:
Could you be more specific @lsegal? What is wrong with the changelog formatting?

benlangfeld:
Took a wild guess; hope that's what you were after :)

lsegal:
LGTM!

lsegal:
Thanks for the PR, @benlangfeld!

PR #1041, opened 2016-12-13
Use RuboCop::RakeTask in Rakefile

olleolleolle:
This PR uses the RakeTask class from RuboCop to define its Rake task.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -36,9 +36,8 @@ rescue LoadError
   nil # noop
 end
 
-task :rubocop do
-  sh "rubocop"
-end
+require 'rubocop/rake_task'
+RuboCop::RakeTask.new
 
 task :default => [:rubocop, :spec]
 


lsegal:
Is there a reason this is necessary?

olleolleolle:
This will run in the same process which could perhaps make it slightly quicker?

lsegal:
Marking this as closed. Depending on the rubocop task brings in a dependency to the Rakefile, which means you need it installed even if you're not running `rake lint`. That seems fine in theory if you use Bundler, but Bundler is not required for yard, and so assuming all dependencies are installed is not valid for this project. For all other requires, you'll notice we have LoadError fallbacks. We could do this here, but that would be equivalent to just running rubocop in a shell. 

The minor performance boost of running in the same process (I don't see any data here to believe there is any significant boost) is outweighed by the coupling issue.

If this was rewritten to use a LoadError fallback it could be considered.

PR #1045, opened 2016-12-27
Cleanup yard specs and fix lots of warnings

ivoanjo:
# Description

A late holidays gift! I wanted to finish this a few days ago but yeah, it took some work :smiley: 

This PR configures a number of recommended rspec features which are off by default so as to preserve backwards compatibility, and enables printing of ruby warnings during test execution. It then goes after many of the warnings being printed, in an effort to fix them all.

I didn't have the opportunity to fix all the warnings, but only a handful of them remain (that tend repeat a bit, thus causing a somewhat noisy output). I left warnings enabled by default so as to motivate others to help out, but this can be easily disabled again :smile: 

I recommend reviewing this PR commit-by-commit, rather than the whole diff at once as that doesn't make a lot of sense. If you're against any of the warning fixes, I can try to fix in another way or remove that fix entirely.

Thanks, and I hope this contribution is relevant!

# Completed Tasks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ Gemfile.lock
 /coverage
 /.bundle
 /vendor
+spec/examples.txt
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,2 @@
 --color
+--require spec_helper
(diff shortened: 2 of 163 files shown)


lsegal:
This is so great! Thank you so much @ivoanjo!

ivoanjo:
It was my pleasure :smile: :+1:

PR #1056, opened 2017-01-12
Enable groups for constants (Resolves #610)

### LINKED ISSUE #610: @!group / @!endgroup doesn't work on constants
example -> 

```
# @!group status

# @api public
UNPUBLISHED = 0
# @api public
PUBLISHED = 1
# @api public
# final - accepted
ACCEPTED = 2
# @api public
# straight up declined
# @api public
DECLINED = 3
# @api public
# waiting for more information
ACCEPTED_PENDING = 4
# @api public
# errors are less than 0
# esign document generation failed - proposal publishing cancelled
ESIGN_FAILURE = -1
# @api public
# waiting for esignature service to finish before publishing
ESIGN_DOCUMENT_IN_PROGRESS = 5

# @!endgroup status
```

result => 
result

an example of where this might be useful is where a group of constants apply to a single attribute on the object

Kjarrigan:
# Description

I started contributing to gosu this month and while working on some issues there I stumbled upon a huge list of constants in the docs and it would be really helpfull if we could group and collapse them. @jlnr found an old issue #610 which stated that this is a feature I have to implement on my own. So this is my first approach. Its definitely not finished yet, but the easy part is done and its functional but not that pretty yet. So I would like to get some feedback if my basic idea was right and if I can look further into this to make it pretty.

# Screenshots

const_summary_expanded
const_summary_collapsed

# ToDo

* [x] Change the stylesheet to make the folded list look like the folded method list (all names in colored boxes, no "linebreak" after each name)
* [x] Add a test

# Completed Tasks

### DIFF
--- a/templates/default/fulldoc/html/js/app.js
+++ b/templates/default/fulldoc/html/js/app.js
@@ -120,6 +120,41 @@ function summaryToggle() {
   } else { localStorage.summaryCollapsed = "expand"; }
 }
 
+function constantSummaryToggle() {
+  $('.constants_summary_toggle').click(function(e) {
+    e.preventDefault();
+    localStorage.summaryCollapsed = $(this).text();
+    $('.constants_summary_toggle').each(function() {
+      $(this).text($(this).text() == "collapse" ? "expand" : "collapse");
+      var next = $(this).parent().parent().nextAll('dl.constants').first();
+      if (next.hasClass('compact')) {
+        next.toggle();
+        next.nextAll('dl.constants').first().toggle();
+      }
+      else if (next.hasClass('constants')) {
+        var list = $('<dl class="constants compact" />');
+        list.html(next.html());
+        list.find('dt').each(function() {
+          $(this).addClass('summary_signature');
+          $(this).text($(this).text().split('=')[0]);
+        });
+        // Add the value of the constant as "Tooltip" to the summary object
+        list.find('pre.code').each(function() {
+          console.log($(this).parent());
+          $(this).parent().prev().attr('title', $(this).text());
+        });
+        list.find('.docstring, .tags, dd').remove();
+        next.before(list);
+        next.toggle();
+      }
+    });
+    return false;
+  });
+  if (localStorage.summaryCollapsed == "collapse") {
+    $('.constants_summary_toggle').first().click();
+  } else { localStorage.summaryCollapsed = "expand"; }
+}
+
 function generateTOC() {
   if ($('#filecontents').length === 0) return;
   var _toc = $('<ol class="top"></ol>');
@@ -241,6 +276,7 @@ $(document).ready(function() {
   searchFrameButtons();
   linkSummaries();
   summaryToggle();
+  constantSummaryToggle();
   generateTOC();
   mainFocus();
 });
--- a/templates/default/fulldoc/html/js/app.js
+++ b/templates/default/fulldoc/html/js/app.js
@@ -135,13 +135,21 @@ function constantSummaryToggle() {
         var list = $('<dl class="constants compact" />');
         list.html(next.html());
         list.find('dt').each(function() {
-          $(this).addClass('summary_signature');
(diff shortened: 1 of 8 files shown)


templates/default/fulldoc/html/js/app.js
```diff
@@ -120,6 +120,38 @@ function summaryToggle() {
...
+      $(this).text($(this).text() == "collapse" ? "expand" : "collapse");
+      var next = $(this).parent().parent().nextAll('dl.summary').first();
+      if (next.hasClass('compact')) {
+        next.toggle();
+        next.nextAll('dl.summary').first().toggle();
+      }
+      else if (next.hasClass('summary')) {
+        var list = $('<dl class="constants summary compact" />');
+        list.html(next.html());
+        list.find('.docstring').remove();
```
lsegal: Why are individual elements being removed here (including elements below). Why not just use CSS to style this?

lib/yard/code_objects/constant_object.rb
```diff
@@ -12,5 +12,12 @@ class ConstantObject < Base
     def value=(value)
       @value = format_source(value)
     end
+
+    # The scope of the constants is always +:class+
+    #
+    # @return [Symbol] +:class+
+    def scope
```
lsegal: I'm not sure how this method is being used. Is there a reason this was added?

Kjarrigan: If there are no groups, but you have constants it raises an exception `NoMethodError: undefined method `scope'` like it did with the classvariable in the failing test altough I'm not sure why this is triggered but I'll definitely look into it.

templates/default/fulldoc/html/js/app.js
Kjarrigan: The current implementation of the method-summaries remove some tags as well so I just copied it but I think your right.

Kjarrigan:
Hi @lsegal, thanks for your feedback. Wasn't sure if my approach would fit into the lib as it's quite some code and I had some problems figuring out where to put my stuff.  But I'll now go on and fix the tests, add a new one for the constant groups and look into your comments.

lib/yard/code_objects/constant_object.rb
lsegal: This probably happens because of some shared logic in the templates. You may want to change the code in the template itself rather than adding new APIs. I'm not sure a #scope method makes sense for a ConstantObject.

Kjarrigan:
_blow the dust away_

Sorry for the long break but I finally looked into this PR again and this time finished it. I reworked the logic, fixed the extisting tests and added a new one. So I guess @lsegal I'm ready for another review.

Attached are two screenshots with a collapsed and an expanded view. If the constants are collapsed the value is shown as tooltip as seen in the first screenshot.

PS: As far as I can tell the failing test on ruby-head is unrelated to this PR.
collapsed
expanded

Kjarrigan:
Sleeping over it I remembered that I didn't test/considered deprecated constants. They work as well but I thought that it would be cool if you can see some indication in the collapsed mode as well - so I added a bit of js and css to make it red and add the Word "Deprecated." in the tooltip as well.

See the attached image.

As far as I can tell there are no tests for the Javascript?
deprecated_const

Kjarrigan:
Rebased and everything green now.

Kjarrigan:
Is this repo still alive?

lsegal:
Not dead, just slower, lately. I will likely make some time on this at some point this month.

lsegal:
Thanks for the hard work on this, @Kjarrigan!

DanRathbun:
When can we expect a new release version ?

Kjarrigan:
@lsegal Thanks for the merge!

thomthom:
The grouping is done by `@!group` and `@!endgroup`, right? Is this working with the C parser? (I'm not seeing any groups - but not sure if I added the tags properly in my source...

PR #1057, opened 2017-01-13
Fix Ruby 2.4 deprecations

nomoon:
# Description

In Ruby 2.4, `::TRUE` & `::FALSE` constants, as well as the `::Bignum` and `::Fixnum` classes are deprecated. The deprecated boolean constants appear on every run of Yard, while the Integer unification-based deprecations only appear in Yard's tests. This fixes both.

# Completed Tasks

### DIFF
--- a/spec/code_objects/constants_spec.rb
+++ b/spec/code_objects/constants_spec.rb
@@ -66,7 +66,9 @@ RSpec.describe YARD::CodeObjects do
   describe :BUILTIN_CLASSES do
     it "includes all base classes" do
       bad_names = []
-      YARD::CodeObjects::BUILTIN_CLASSES.each do |name|
+      builtin_classes = YARD::CodeObjects::BUILTIN_CLASSES
+      builtin_classes -= ["Bignum", "Fixnum"] if 0.class == Integer
+      builtin_classes.each do |name|
         begin
           bad_names << name unless eval(name).is_a?(Class)
         rescue NameError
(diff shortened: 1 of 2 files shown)


spec/code_objects/constants_spec.rb
```diff
@@ -66,7 +66,9 @@
   describe :BUILTIN_CLASSES do
     it "includes all base classes" do
       bad_names = []
-      YARD::CodeObjects::BUILTIN_CLASSES.each do |name|
+      builtin_classes = YARD::CodeObjects::BUILTIN_CLASSES
+      builtin_classes -= ["Bignum", "Fixnum"] if 0.class == Integer
```
lsegal: This change means we are no longer testing for all builtin classes. This should probably be wrapped in a version check, but at best, we should find a way to test this without emitting a warning, if possible.

nomoon: As far as I'm aware, any access of `::Bignum` or `::Fixnum` will emit a deprecation warning in Ruby 2.4 (as in the `val(name).is_a?(Class)` at line 73 of `constants_spec.rb` above). The new code only omits those two classes in situations where that would be the case (where `0.class == Integer`, which is only the case from 2.4 onwards).

lsegal: I'm going to decline this change and cherry-pick 3bfa1b2 only. In this case, silencing Ruby's deprecation is preferable (via `$VERBOSE = nil`)

Issue #1062, opened 2017-01-19
`bundler exec rake spec` failes with undefined method `shared_context_metadata_behavior='


thomthom:
I had pulled the latest yard source from github. Ran `bundler install` and then tried to run `bundler exec rake spec`, but I'm getting errors in return. 

## Steps to reproduce

This is the minimal reproduction for the issue. I've done my best to remove
all extraneous code and unique environment state on my machine before providing
these steps:

1. Clone repo
2. Run `bundler install`
3. Run `bundler exec rake spec`

## Actual Output

```
C:\Users\tthomas2\SourceTree\yard (master)
λ bundle exec rake spec
C:/Ruby23/bin/ruby.exe -I'C:/Ruby23/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/lib';'C:/Ruby23/lib/ruby/gems/2.3.0/gems/rspec-support-3.4.1/lib' 'C:/Ruby23/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/exe/rspec' --pattern 'spec/**{,/*/**}/*_spec.rb'
C:/Users/tthomas2/SourceTree/yard/spec/spec_helper.rb:159:in `block in <top (required)>': undefined method `shared_context_metadata_behavior=' for #<RSpec::Core::Configuration:0x310fab8> (NoMethodError)
        from C:/Ruby23/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/lib/rspec/core.rb:97:in `configure'
        from C:/Users/tthomas2/SourceTree/yard/spec/spec_helper.rb:119:in `<top (required)>'
        from C:/Ruby23/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/lib/rspec/core/configuration.rb:1295:in `require'
        from C:/Ruby23/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/lib/rspec/core/configuration.rb:1295:in `block in requires='
        from C:/Ruby23/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/lib/rspec/core/configuration.rb:1295:in `each'
        from C:/Ruby23/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/lib/rspec/core/configuration.rb:1295:in `requires='
        from C:/Ruby23/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/lib/rspec/core/configuration_options.rb:109:in `block in process_options_into'
[...]

PR #1063, opened 2017-01-19
Allowing YARD to parse content of C++ namespaces. Fixes #809.

### LINKED ISSUE #809: C-Parser: cant find method definition when using namespaces
SampleCode

```c++
VALUE rb_cWXRect;

namespace RubyWX {
namespace Rect {

VALUE _alloc(VALUE self)
{
	return wrap(new wxRect());
}

/*
 * call-seq:
 *   inspect -> String
 *
 * Human-readable description.
 * ===Return value
 * String
*/
VALUE _inspect(VALUE self)
{
	return rb_sprintf( "%s(%d, %d, %d, %d)",
		rb_obj_classname( self ),
		FIX2INT(_getX(self)),
		FIX2INT(_getY(self)),
		FIX2INT(_getWidth(self)),
		FIX2INT(_getHeight(self)));
}

}
}

void Init_WXRect()
{
	using namespace RubyWX::Rect;
	rb_cWXRect = rb_define_class("Rect",rb_cObject);

	rb_define_alloc_func(rb_cWXRect,_alloc);

	rb_define_method(rb_cWXRect,"inspect",RUBY_METHOD_FUNC(_inspect),0);

}
```

returns this
```
Files:           2
Modules:         0 (    0 undocumented)
Classes:         1 (    1 undocumented)
Methods:         1 (    1 undocumented)
 0.00% documented
```
currenty the Files2 is a bug that happens because one of the o.file is nil
but for my self i cant fix it that it does find the method with the c++ namespaces

thomthom:
# Description

This Pull Request attempts to address #809 - where content inside of C++ namespaces isn't processed. This is an issue for source code that use the Ruby C API in a C++ environment.

What this fix do is simply make the C Parser not treat the content of namespaces as body statements. Instead it just ignores the namespaces all together.

I'm not sure if it's perfect, but we're working on getting a very sizable amount of code parsed so I'll follow up with improvements as needed if we run into them.

# Completed Tasks

### DIFF
--- a/lib/yard/parser/c/c_parser.rb
+++ b/lib/yard/parser/c/c_parser.rb
@@ -38,6 +38,7 @@ module YARD
             when '#'; consume_directive
             when '/'; consume_comment
             when /\s/; consume_whitespace
+            when '}'; advance # Skip possible C++ namespace closing brackets.
             else consume_toplevel_statement
             end
           end
@@ -75,6 +76,8 @@ module YARD
           line = @line
           decl = consume_until(/[{;]/)
           return nil if decl =~ /\A\s*\Z/
+          # Skip C++ namespace - treat content as top level statement.
+          return nil if decl =~ /\A(namespace)/
           statement = ToplevelStatement.new(nil, @file, line)
           @statements << statement
           attach_comment(statement)
--- a/spec/parser/c_parser_spec.rb
+++ b/spec/parser/c_parser_spec.rb
@@ -26,6 +26,29 @@ RSpec.describe YARD::Parser::C::CParser do
       end
     end
 
+    describe "C++ namespace" do
+      before(:all) do
+        file = File.join(File.dirname(__FILE__), 'examples', 'namespace.cpp.txt')
+        parse(File.read(file))
+      end
+
+      it "parses Rect class" do
+        obj = YARD::Registry.at('Rect')
+        expect(obj).not_to be nil
+        expect(obj.docstring).not_to be_blank
+      end
+
+      it "parses method inside of namespace" do
+        obj = YARD::Registry.at('Rect#inspect')
+        expect(obj.docstring).not_to be_blank
+      end
+
+      it "parses method after namespace" do
+        obj = YARD::Registry.at('Rect#hello_world')
+        expect(obj.docstring).not_to be_blank
+      end
+    end
+
     describe "Source located in extra files" do
       before(:all) do
         @multifile = File.join(File.dirname(__FILE__), 'examples', 'multifile.c.txt')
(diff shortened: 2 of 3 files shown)


thomthom:
I had to add another commit to correct a last moment issue that made RuboCop fail. Do I have to go back and squash this? Or can you just use the GitHub Merge and Squash feature?

I was at first not running rubocop before I submitted because I threw errors on all files due to line endings. That's because git normalize to OS line endings upon checkout - but then correct them when you check in.

I ended up adding these lines temporarily to `.rubocop` to mute those warnings:

```
EndOfLine:
  Enabled: false
```

Since git normalize line ending upon checking, is that RuboCop check really needed?

thomthom:
hmm... hold this PR. I might have run into a regression. Let me run more tests to be sure.

Gemfile
```diff
@@ -2,7 +2,7 @@
 source 'https://rubygems.org'
 
 group :development do
-  gem 'rspec', '~> 3.0'
+  gem 'rspec', '~> 3.5'
```
thomthom: Ah, so I had not intended this to be part of this PR. But this was something I had to do in order to get things running. See #1062

lib/yard/parser/c/c_parser.rb
```diff
@@ -38,6 +38,7 @@ def parse_toplevel
             when '#'; consume_directive
             when '/'; consume_comment
             when /\s/; consume_whitespace
+            when '}'; advance # Skip possible C++ namespace closing brackets.
```
thomthom: So this one, I'm not sure if you'll like this much. I can try to refactor it if you prefer. Though any suggestions would be appreciated. The issue here was that when there are namespaces my fix was to just ignore it. But then `consume_toplevel_statement` would later run into stray `}` (closing namespaces) that would throw it off.

So this line is currently very generic in advancing over any `}` in the top level. I could try to keep track of the depth of the C++ namespaces and only advance over `}` a matching number of times.

It all depend on how low impact we want to keep this. I ran this fix over our code base with over 1300 methods and 98 classes - no documentation objects where lost.

Gemfile
lsegal: Undo this change and rebase off of master since this fix was pushed separately.

thomthom: Ok. So when rebasing, do I then create a new PR? since that changes the history of the branch. (I'm not that familiar with rebase.)

lib/yard/parser/c/c_parser.rb
lsegal: I'm not too concerned, since the worst case is that it continues past a toplevel element to find the next one.

Issue #1067, opened 2017-01-31
Regex Duplicate Range warning in docstring_parser.rb


cschneid:
The regex in docstring_parser.rb contains overlapping character classes, which is emitting a warning from Ruby.

The warning:

```
/Users/cschneid/.rvm/gems/ruby-2.2.5/gems/yard-0.9.8/lib/yard/docstring_parser.rb:119: warning: character class has duplicated range: /\A[\r\n\s]+|[\r\n\s]+\Z/
```

From the Ruby docs, the \s class handles \r and \n itself, so the explicit \r and \n parts of the line aren't necessary.

```
/\s/ - A whitespace character: /[ \t\r\n\f]/
```

## Steps to reproduce

Require the file with warnings enabled.

```
› irb -w
>> require 'yard/docstring_parser'
/Users/cschneid/.rvm/gems/ruby-2.2.5/gems/yard-0.9.8/lib/yard/docstring_parser.rb:119: warning: character class has duplicated range: /\A[\r\n\s]+|[\r\n\s]+\Z/
=> true
```

## Expected Output

No warnings.

## Environment details:

* OS: OSX
* Ruby version (`ruby -v`): ruby 2.2.5p319 (2016-04-26 revision 54774) [x86_64-darwin15]

* YARD version (`yard -v`): yard 0.9.8
* Relevant software dependency/versions: None.

I have read the [Contributing Guide][contrib].

[contrib]: https://github.com/lsegal/yard/blob/master/CONTRIBUTING.md

PR #1068, opened 2017-01-31
Remove overlapping character classes to eliminate a warning

cschneid:
# Description

\s already covers the \r and \n characters, so Ruby was emitting a
warning about overlapping characters.

#1067 details the issue.

I didn't change this to just use `.strip`, since I wasn't sure if there was a reason that a regex & gsub were being used. It's possible that `.strip` from the stdlib would be clearer.

# Completed Tasks


cschneid:
Closing, will reopen w/ a strip version.

PR #1069, opened 2017-02-01
Use String#strip, rather than regex

cschneid:
The previous regex was emitting a warning about overlapping character
classes, and the regex was complicated. Use the stdlib's strip function
since it's more clear, and also doesn't emit warnings.

# Description

An updated PR for #1068 

# Completed Tasks

* [x] Git commits have been cleaned up (squash WIP / revert commits). **this is why I have a new PR here**


cschneid:
This fixes #1067

PR #1075, opened 2017-02-25
API docs for CLI: avoid misrendered dashes

olleolleolle:
This change avoids rendering `--` as a single emdash character in the generated HTML.

By pointing at the `yard help doc` invocation, we side-step the issue altogether.


lsegal:
The doc fix here is fine, but it seems like there's an underlying rendering bug that needs to be fixed as well.

PR #1083, opened 2017-04-09
Add gem removal hook to remove YARD and yri data on uninstalling gems

KitaitiMakoto:
# Description

When you uninstall some gems, YARD and yri data should be removed as well as RDoc data(see 
[here][rdoc]).

[rdoc]: https://github.com/ruby/ruby/blob/68c9c97f139c8e4f0543a3f431fdb567ad6b1ac3/lib/rubygems/uninstaller.rb#L271

# Completed Tasks

### DIFF
--- a/lib/yard/rubygems/hook.rb
+++ b/lib/yard/rubygems/hook.rb
@@ -72,6 +72,14 @@ module YARD
     end
 
     ##
+    # Pre uninstalls hook that removes documentation
+    #
+
+    def self.removal_hook(uninstaller)
+      new(uninstaller.spec).remove
+    end
+
+    ##
     # Loads the YARD generator
 
     def self.load_yard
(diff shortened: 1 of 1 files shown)


lsegal:
Does RDoc do this?

KitaitiMakoto:
RDoc gem itself doesn't.
But rubygems lib bundled with Ruby removes RDoc's data on uninstalling:
https://github.com/ruby/ruby/blob/68c9c97f139c8e4f0543a3f431fdb567ad6b1ac3/lib/rubygems/uninstaller.rb#L271

As we see the result, by uninstalling gems, we remove documentation files. Don't you think it's nice if there is no documentation files including YARD's after uninstalling gems?

lsegal:
I'm fine with this, but I want to make sure we're doing what other tools are.

KitaitiMakoto:
Thank you so much!

PR #1087, opened 2017-04-21
Add option to restrict gems from Gemfile (deps only)

tonobo:
# Description

It will add an new feature which basically depends on `--gemfile` option.  
The new `--gemfile-deps` option will only render **specified** Gems, instead of rendering the whole Gemfile.lock (automatically resolved dependencies) as `--gemfile` does.

# Completed Tasks

### DIFF
--- a/lib/yard/cli/server.rb
+++ b/lib/yard/cli/server.rb
@@ -133,13 +133,26 @@ module YARD
         end
       end
 
-      def add_gems_from_gemfile(gemfile = nil)
+      def add_gems_from_gemfile(gemfile = nil, dependencies_only = false)
         require 'bundler'
         gemfile ||= "Gemfile"
         if File.exist?("#{gemfile}.lock")
-          Bundler::LockfileParser.new(File.read("#{gemfile}.lock")).specs.each do |spec|
-            libraries[spec.name] ||= []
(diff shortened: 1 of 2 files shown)


lsegal:
I still don't see a problem statement here. Why is this needed?

dopykuh:
I have a lot of gems which i will automatically generate documentation for, but i don't care about the resolved depencencies. I just want a list of my libaries not tens of others.

lsegal:
I understand this now. I'd prefer a better approach than adding yet another `--gemfile*` argument that does mostly the same thing as the other with an unclear distinction.

Calling them "dependencies" is not clear. Transitive dependencies are also dependencies, so saying "only dependencies" should not mean anything different. "Ignore transitive dependencies" would be more clear, but that's obviously wordy.

I'd actually be open to changing the behavior of `--gemfile` slightly to pick direct dependencies if `Gemfile` is selected, and include _all_ dependencies if the `Gemfile.lock` is selected. This is technically a breaking change, though I believe this is the behavior that most users will want anyway, so breaking is relative. One could call the inclusion of transitive dependencies a bug in its own right.

Doing that would definitely mean waiting on any feedback though, and some investigation to see how many people might be affected.

lsegal:
> [!NOTE]

PR #1092, opened 2017-05-08
fix duplication of return tags in C

yagisumi:
# Description

fix duplication of return tags on methods end with a question mark in C

# Completed Tasks


lsegal:
@yagisumi 

Hi,

Thanks for the pull request!

It doesn't look like this PR has tests. This probably needs tests to ensure we're doing the right thing. Can you add them?

Thanks again!

PR #1095, opened 2017-05-26
Allow "@param foo" to document a _foo parameter

mvidner:
# Description

Consider this example of an abstract base and a concrete implementation:

```ruby
class Base
  # This method is very important and here is its description.
  # @param thing [Foo] the thing to process
  def process(_thing)
    raise NotImplementedError,
      "Subclasses of #{Module.nesting.first} must override #{__method__}"
  end
end

class Concrete < Base
  # (see Base#process)
  def process(thing)
    thing.foo
  end
end
```

The base class documents the *thing* parameter but it does not use it.
Accordingly, the parameter is renamed to `_thing` (which makes RuboCop happy).

But YARD gives a warning "@param tag has unknown parameter name: thing" for `Base`.
If we used "@param _thing" then an analogous warning would be given for `Concrete` (which uses the `(see Class#method)` shortcut for docs copying).

Therefore I propose this change, which allows `@param foo` to match a parameter named `_foo`.

(For reference, the actual code from my team's work affected by this is: [source](https://github.com/config-files-api/config_files_api/blob/7f411362acbecd8d723ba2dd730a012c9f766be0/lib/cfa/placer.rb#L6), [rendered docs](http://www.rubydoc.info/github/config-files-api/config_files_api/CFA/Placer#new_element-instance_method))

# Completed Tasks

### DIFF
--- a/lib/yard/docstring_parser.rb
+++ b/lib/yard/docstring_parser.rb
@@ -278,6 +278,8 @@ module YARD
       next unless parser.object.is_a?(CodeObjects::MethodObject)
       next if parser.object.is_alias?
       names = parser.object.parameters.map {|l| l.first.gsub(/\W/, '') }
+      alt_names = names.find_all {|n| n.start_with?("_") }.map {|n| n[1..-1] }
+      names.concat(alt_names)
       seen_names = []
       infile_info = "\n    in file `#{parser.object.file}' " \
                     "near line #{parser.object.line}"
--- a/spec/docstring_parser_spec.rb
+++ b/spec/docstring_parser_spec.rb
@@ -215,6 +215,14 @@ eof
       eof
     end
 
+    it "does not warn about unused named parameters" do
+      expect(log).to_not receive(:warn).with(/@param tag has unknown parameter name: bar/)
+      YARD.parse_string <<-eof
+        # @param bar useful in subclasses
+        def foo(_bar) end
+      eof
+    end
+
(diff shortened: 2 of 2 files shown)


lsegal:
I'm not sure this can be supported. The warning here is correct that the parameter does not match any valid parameter name in source. Changing this expectation would be a breaking change for downstream tooling that would now be expected to match parameter names against `foo` or `_foo`. This also opens the door for other arbitrary rule changes down the road that YARD is now expected to work around. It doesn't seem reasonable to me to expect YARD's naming rules to be coupled to conventions defined by a separate linting tool-- especially if we're not affording that same responsibility the other way around.

Given that this warning doesn't actually affect how HTML generates (in your use case), hiding this doesn't actually do anything to solve the root problem here: that YARD's reftag syntax is not robust enough to support this naming convention.

In reality, the best solution here is to use an `@overload` tag to rewrite your signature if you really want reliable docs that don't confuse your user. The code below would work without a warning, and it also has the side benefit of also displaying your signature correctly in rendered docs.

```ruby
# @overload process(thing)
#   @param thing [String] the thing
def process(_thing)
  # whatever you like in here
end
```

mvidner:
> The code below would work without a warning, and it also has the side benefit of displaying your signature correctly in rendered docs.

Oh, that works! Thank you!

> ... reftag ...

Ah, you mean http://www.rubydoc.info/gems/yard/file/docs/GettingStarted.md#Reference_Tags , I'll look more into that too.

PR #1098, opened 2017-06-03
allow aliases of class methods in C

yagisumi:
# Description

To allow aliases of class methods in C.

# Completed Tasks

### DIFF
--- a/lib/yard/handlers/c/alias_handler.rb
+++ b/lib/yard/handlers/c/alias_handler.rb
@@ -1,16 +1,26 @@
 # frozen_string_literal: true
 class YARD::Handlers::C::AliasHandler < YARD::Handlers::C::Base
-  MATCH = /rb_define_alias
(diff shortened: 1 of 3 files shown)


lsegal:
I'm okay with this as a stopgap but it looks like this only solves for one specific usage where `rb_singleton_class()` is used directly. This wouldn't account for:

```c
VALUE foometa = rb_singleton_class("Foo");
rb_define_alias(foometa, "a", "b");
```

In fact, it seems as though `rb_singleton_class()` can be used as an argument to any of the `rb_def*` calls, and if so, it would be smarter to step back and look at the way `namespace_for_variable` resolves values.

yagisumi:
Hi. I think so, too. 
Anyway, I wanted to solve common patterns.
For now, I am glad if handle_alias method has a scope argument.
Then I can write handlers for my documentation.

PR #1099, opened 2017-06-23
Document block for Struct.new if present

akihikodaki:
# Description
The documentation of Ruby says:
> If a block is given it will be evaluated in the context of StructClass, assing the created class as a parameter:
> (snip)
> This is the recommended way to customize a struct. Subclassing an anonymous struct creates an extra anonymous class that will never be used.

This change implements documentation for such usage.

# Completed Tasks

* [X] I wrote tests and ran `bundle exec rake` locally (if code is attached to PR).


lsegal:
Thanks for the PR!

PR #1104, opened 2017-07-17
Parse ary and new_ary types in call-seq

ypresto:
# Description

Lastest Ruby's array.c contains `ary` and `new_ary` return types which is not handled by yard for now. So `@return` for `Array#map` is missing from generated yard doc. This PR fixes it..!

# Completed Tasks


lsegal:
This seems to fail tests on the `&.` syntax, which we can't use.

ypresto:
Oh, I'll rewrite it with `try()` instead!

ypresto:
fixed!

lsegal:
Thanks for the PR!

PR #1168, opened 2018-05-19
Add support for quoted symbols

cboos:
# Description

Constants containing *quoted symbols* can now be documented correctly.

Prior to the fix, the value `:"sym+bol"` would be shown as `sym+bol`
and worse, the empty symbol (`:''`) would have triggered an error in
`MethodHelper#format_constant`, as the corresponding constant value 
was the empty String.

For example, this code:

```ruby
class X
  Y = :''
end
```

would trigger the following error in YARD 0.9.12:

```
[error]: NoMethodError: undefined method `[]' for nil:NilClass
[error]: Stack trace:
        .../lib/yard/templates/helpers/method_helper.rb:69:in `format_constant'
        .../templates/default/module/html/constant_summary.erb:10:in `block in _erb_cache_18'
        .../templates/default/module/html/constant_summary.erb:6:in `each'
        .../templates/default/module/html/constant_summary.erb:6:in `_erb_cache_18'
        .../lib/yard/templates/template.rb:287:in `erb'
        .../lib/yard/templates/template.rb:365:in `render_section'
```	

The `ConstantHandler` didn't treat the `:dyna_symbol` node specially, and
simply calling the source method on that statement would return the characters
making up the symbol, not the actual source code for that symbol (i.e. the prefix `:`
and the wrapping quotes).

This is fixed in commit 5b1dc8ae.

Besides, when investigating the issue, I first followed a wrong track by fixing
an error in the tokenisation code of the parser. This proved to be not
necessary to fix my original error, but I think the change will not harm
so I've left it in the PR as a separate commit (db005970).

Some more details about this other fix: in RubyParser, 
a sequence of tokens gets built, and a  `[:symbeg, :const]` pair 
gets replaced by a `:symbol` token.
However, there are actually three possibilities (maybe more?)
for constructing a Symbol:

 1. `:symbeg`, `:const`  (e.g. `:BAR`)
 2. `:symbeg`, `:tstring_content`, `:tstring_end` (e.g. `:"B+z"`)
 3. `:symbeg`, `:tstring_end` (e.g. `:''`)

So I added support for case 2. and 3. as well.

# Completed Tasks

### DIFF
--- a/lib/yard/handlers/ruby/constant_handler.rb
+++ b/lib/yard/handlers/ruby/constant_handler.rb
@@ -20,7 +20,11 @@ class YARD::Handlers::Ruby::ConstantHandler < YARD::Handlers::Ruby::Base
 
   def process_constant(statement)
     name = statement[0].source
-    value = statement[1].source
+    value = if statement[1].type == :dyna_symbol
+              statement[1].first.source.intern.inspect
+            else
+              statement[1].source
+            end
     obj = P(namespace, name)
     if obj.is_a?(NamespaceObject) && obj.namespace == namespace
       raise YARD::Parser::UndocumentableError, "constant for existing #{obj.type} #{obj}"
--- a/lib/yard/parser/ruby/ruby_parser.rb
+++ b/lib/yard/parser/ruby/ruby_parser.rb
@@ -285,9 +285,10 @@ module YARD
           if token == :symbeg
             @symbol = [:symbol, data, [lineno, charno]]
           elsif @symbol
-            if token == :tstring_content
+            case token
+            when :tstring_content
               @symbol[1] += data
-            elsif token == :tstring_end || token == :const
+            when :tstring_end, :const, :ident
               @symbol[1] += data
               @tokens << @symbol
               @symbol = nil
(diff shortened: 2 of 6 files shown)


cboos:
Sorry, I did run 'rake rubocop' and 'rake spec', but forgot to run 'rake yard': now the latter fails. I'll investigate and provide a follow-up.

cboos:
Btw, as both the contributing guide and the PR template advise to do a `bundle exec rake`, what about including `:yard` as an extra step to the default build? Running YARD on itself is a nice last validation step on top of the tests!

lsegal:
Thank you for the PR.

I agree that running yard as part of the rake might be worthwhile, but I think we would need some kind of way to validate the results. Without validating, I'm not sure how this would work.

One thing we could do is integration test the `yard doc` command as a separate RSpec test and just smoke test for a valid exit code (0). That would be a reasonable test and would fit nicely into our testing infrastructure. I would accept a PR for this.

cboos:
PR #1171 added a smoke test to run yard on yard's source code during CI.

PR #1175, opened 2018-07-04
Change some type annotation in YARD doc

gmalette:
# Description

Some of the type annotations in the documentation referrenced types that
were not in scope for the annotation. This means that a compiler
generating the doc could incorrectly associate the constants with
another type which would be in scope.

# Completed Tasks

* ~~[ ] I wrote tests and ran `bundle exec rake` locally (if code is attached to PR).~~ (no code)

### DIFF
--- a/lib/yard/handlers/base.rb
+++ b/lib/yard/handlers/base.rb
@@ -181,7 +181,7 @@ module YARD
         # the handlers, otherwise the same code will be parsed
         # multiple times and slow YARD down.
         #
-        # @param [Parser::RubyToken, Symbol, String, Regexp] matches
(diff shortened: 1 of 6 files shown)


lib/yard/handlers/base.rb
```diff
@@ -181,7 +181,7 @@ def inherited(subclass)
         # the handlers, otherwise the same code will be parsed
         # multiple times and slow YARD down.
         #
-        # @param [Parser::RubyToken, Symbol, String, Regexp] matches
+        # @param [Parser::Ruby::Legacy::RubyToken, Symbol, String, Regexp] matches
```
gmalette: I'm not 100% certain this is the right one, but it's the only `RubyToken` I could find in this gem.

PR #1178, opened 2018-07-20
Fix annotation in CodeObjects::Base#files

gmalette:
The original annotation incorrectly was Array<String> but the format is

```
[ [file, line] ]
```

Therefore the annotation should be `Array<Array<String, Integer>>`,
given there is no prescribed way to describe tuples

# Description

# Completed Tasks

### DIFF
--- a/lib/yard/code_objects/base.rb
+++ b/lib/yard/code_objects/base.rb
@@ -132,7 +132,7 @@ module YARD
     # @see NamespaceMapper.register_separator
     class Base
       # The files the object was defined in. To add a file, use {#add_file}.
-      # @return [Array<String>] a list of files
+      # @return [Array<Array<String, Integer>>] a list of files
       # @see #add_file
       attr_reader :files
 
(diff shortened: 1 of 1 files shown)


lib/yard/code_objects/base.rb
```diff
@@ -132,7 +132,7 @@ def push(value)
     # @see NamespaceMapper.register_separator
     class Base
       # The files the object was defined in. To add a file, use {#add_file}.
-      # @return [Array<String>] a list of files
+      # @return [Array<Array<String, Integer>>] a list of files
```
lsegal: You can use `Array<Array(String, Integer)>` to denote ordered tuples. Try it out at https://yardoc.org/types.html

gmalette: TIL! Thanks, fixed 😄

gmalette: Given `X<A, B>` means `X parameterized by either A or B`, how do you annotate `X parameterized by both A and B` ?

lsegal:
Thanks!

PR #1185, opened 2018-09-04
Stable sort tags in Docstring#to_raw.

thomthom:
# Description

My team is maintaining a large Ruby API for our application. It's all written in C/C++. To aid developers we have been generating Ruby stubs of the API which IDEs can use to provide code insight and auto-complete. For that we've used YARD, via a custom template.

But we have had some issues where the `@param` tags got reordered.

I found the culprit in `Docstring.to_raw`. More specifically `tags.sort_by(&:tag_name)`. Because Ruby's sorting methods aren't stable (https://stackoverflow.com/questions/15442298/is-sort-in-ruby-stable) tags of the same type might end up being re-ordered.

(I was using `Docstring.to_raw` because in the process of generating Ruby stubs from the C++ source I had to make some smaller changes for them to work better.)

This change makes the sorting of tags in `Docstring#to_raw` stable.

# Completed Tasks

### DIFF
--- a/lib/yard/docstring.rb
+++ b/lib/yard/docstring.rb
@@ -205,7 +205,7 @@ module YARD
     # @since 0.7.0
     # @todo Add Tags::Tag#to_raw and refactor
     def to_raw
-      tag_data = tags.sort_by(&:tag_name).map do |tag|
+      tag_data = tags.map do |tag|
         case tag
         when Tags::OverloadTag
           tag_text = "@#{tag.tag_name} #{tag.signature}\n"
@@ -271,7 +271,7 @@ module YARD
     # @param [#to_s] name the tag name to return data for, or nil for all tags
     # @return [Array<Tags::Tag>] the list of tags by the specified tag name
     def tags(name = nil)
-      list = @tags + convert_ref_tags
+      list = stable_sort_by(@tags + convert_ref_tags, &:tag_name)
       return list unless name
       list.select {|tag| tag.tag_name.to_s == name.to_s }
     end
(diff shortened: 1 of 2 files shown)


thomthom:
I'll make it private for now. Though I am curious about baking it into `#tags`. But as you say, that's probably better as a separate PR. (I suspect it can snowball with some tests that might need updating.)

thomthom:
Actually, I quickly made a test by ensuring `#tags` always return a stable sorted set of tags. No tests failed. I could update this PR with those change if you want.

If so, do you want the sorting method as a private in docstring, or should I add it to `Enumerable` in `core_ext`?

Here's a generic version I have in docstring.rb right now:

```ruby
    # A stable sort_by method.
    #
    # @param [Enumerable]
    # @return [Array]
    def stable_sort_by(list)
      list.each_with_index.sort_by {|tag, i| [yield(tag), i] }.map(&:first)
    end
```

thomthom:
Here's two variants that both stable sort `Docstring#tags`. The latter extends `Enumerable` with `#stable_sort_by` for general usage.

https://github.com/lsegal/yard/compare/master...thomthom:dev-stable-sort-general
https://github.com/lsegal/yard/compare/master...thomthom:dev-stable-sort-core-ext

lsegal:
I think sticking it in Docstring is better than messing with `core_ext`, the latter being more or less a deprecated approach

thomthom:
Sounds good to me. I avoid modifying core classes in my own projects. Though I wasn't sure what the convention in YARD was.

So do you want me to update this PR with making `#tags` always return a stable sorted set of tags?

lsegal:
I think I would be fine with that, yes.

lsegal:
It looks like Travis is having issues right now. Tried restarting the build but it's failing in a weird way.

lsegal:
Merging since this passes locally.

thomthom:
Will there be a new release soon that include this commit?

PR #1194, opened 2018-09-21
Fix builds

skalee:
# Problem description

Currently, CI builds are failing after running ca. 10% of total number of examples. This happens due to a subtle bug in Yard's documentation introduced in 5deb08c329c0609429c2b8670ff139eb7a446635 (#1185).

Many RSpec examples rely on processing Yard's own documentation. Among them, following one: https://github.com/lsegal/yard/blob/75ca29898d2f6a9b1425125e6db4e42a2f3c2c4b/spec/cli/yardoc_spec.rb#L187-L190 The problem is that running Yard with `--fail-on-warning` option may cause abnormal exit in case of warnings. And this is actually happening, because `@param` tag should be followed by description:
https://github.com/lsegal/yard/blob/5deb08c329c0609429c2b8670ff139eb7a446635/lib/yard/docstring.rb#L380

# Changes

The aforementioned RSpec example has been surrounded with begin-rescue-end block, so it no longer breaks builds in situations like this. Apart from that, a new Rake task `test_doc` has been implemented, which tests Yard's own documentation, and ensures that no warnings are generated. Finally, this Rake test has been included in Rake's default task, so documentation well-formedness is now tested in CI.

The build will fail, but this is expected. Bug introduced in #1185 hasn't been fixed (and I'm not going to fix it because people who know this project well will do it better, probably). Nevertheless, all tests are run now, and the build output is meaningful. Also, Rubies that Travis tests against have been updated.

# Completed Tasks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,10 @@ before_install:
 
 # Specify which ruby versions you wish to run your tests on, each version will be used
 rvm:
-  - 2.2.9
-  - 2.3.6
-  - 2.4.3
-  - 2.5.0
+  - "2.2"
+  - "2.3"
+  - "2.4"
+  - "2.5"
 # - ruby-head
 # - jruby
 
--- a/Rakefile
+++ b/Rakefile
@@ -41,7 +41,12 @@ task :rubocop do
   sh "rubocop"
 end
 
-task :default => [:rubocop, :spec]
+desc "Generate documentation for Yard, and fail if there are any warnings"
+task :test_doc do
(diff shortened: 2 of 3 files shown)


lsegal:
This is great thanks!

skalee:
Sorry, it appears that there are new code style offences which I need to fix first.

skalee:
Style offence has been suppressed. Furthermore, newer Rubies have been added to the build matrix.

lsegal:
Doc fix was pushed to master so build should be passing now. Thanks a ton for diagnosing and fixing this!

PR #1197, opened 2018-10-03
Add Ruby versions back to 2.0

lsegal:
# Description

Testing Ruby 2.0/2.1 on CI.

# Completed Tasks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ before_install:
 
 # Specify which ruby versions you wish to run your tests on, each version will be used
 rvm:
+  - "2.0"
+  - "2.1"
   - "2.2"
   - "2.3"
   - "2.4"
--- a/spec/templates/method_spec.rb
+++ b/spec/templates/method_spec.rb
@@ -101,7 +101,7 @@ RSpec.describe YARD::Templates::Engine.template(:default, :method) do
     it_should_behave_like "all formats"
   end
 
-  describe "method with keyword arguments spalt" do
+  describe "method with keyword arguments" do
     before do
       @template = :method006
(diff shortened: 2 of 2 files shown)


skalee:
After not very detailed investigation, I suppose it may be due to the fact that Ripper in Ruby 2.0's stdlib is unable to parse some pieces of code examples from tests. Try in console:

      Ripper.sexp <<-'eof'
        # @param [String] x the x argument
        # @param [Boolean] y the y argument
        # @param [kword1] keyword 1
        # @param [kword2] keyword 2
        def m(x, y, *args, kword1: 123, kword2:, **) end
      eof

It will return nil in Ruby 2.0, and a non-empty array in Ruby 2.5.

I suppose this issue may be difficult to fix. Possibly, this RMI version is buggy. Therefore, I think it's a good idea to add something like "pending if Ruby < 2.1" to get these builds passing.

lsegal:
Ruby 2.0 didn't support keywords. This works:

```ruby
  Ripper.sexp <<-'eof'
    # @param [String] x the x argument
    # @param [Boolean] y the y argument
    def m(x, y, *args) end
  eof
```

It seems like YARD works but we broke the test and need a separate one for keywords specifically.

lsegal:
Sorry actually Ruby 2.0 didn't support _required_ keyword arguments: https://robots.thoughtbot.com/ruby-2-keyword-arguments

So the following should work too:

```ruby
  Ripper.sexp <<-'eof'
    # @param [String] x the x argument
    # @param [Boolean] y the y argument
    # @param [kword1] keyword 1
    # @param [kword2] keyword 2
    def m(x, y, *args, kword1: 123, **) end
  eof
```

PR #1200, opened 2018-10-13
Markup agnostic extra files

### LINKED ISSUE #1192: Yard's meta data make AsciiDoc render ugly when used outside Yard
Yard supports [meta tags in textual documents](https://github.com/lsegal/yard/blob/master/docs/GettingStarted.md#adding-meta-data-to-extra-files). For instance, one could be annotated as follows:

```
# @title The Document Title
# @author The Author Name
```

The problem is that this special markup breaks output outside Yard, i.e. in GitHub. In Markdown, a regular HTML comment can be used to workaround this problem:

```
<!--
# @title The Document Title
# @author The Author Name
-->
```

However, in AsciiDoc nothing helps.

## Attempt 1: HTML comments

```
<!--
# @title The Document Title
# @author The Author Name
-->
```

All works correct in Yard. However, outside Yard HTML tags are escaped, which results with even uglier output:

zrzut ekranu 2018-09-15 o 10 53 34

## Attempt 2: AsciiDoc comments

```
////
# @title The Document Title
# @author The Author Name
////
```

Document looks good both in Yard and outside it, but document title remains unchanged (equals to file name).

## Attempt 3: AsciiDoc conditional processing

```
ifndef::env-browser,env-github[]
# @title The Document Title
# @author The Author Name
endif::[]
```

Document looks good in Yard, but its title remains unchanged.

skalee:
# Description

Rewrite how comments are handled in textual files (extra objects).  Prefer markup-agnosticism over sticking to HTML comments.  Do not remove comment tags, as they are bound to specific markup, and hidden from user's eye anyway. As an effect, more types of comments are now supported, including ones from AsciiDoc and Textile. Fixes #1192.

This is a somewhat breaking change (see).

# Possible improvements

If required, I can restore previous behaviour changed in dc5788a, making this change a non-breaking one. Well, at least in terms of how it was formally specified in specs.

It's just a matter of adding following lines to the altered `case` construct:

```ruby
when /^\s*$/
  cut_index_end = index
  break
```

# See also

- discussion in [Yard's mailing list](https://groups.google.com/forum/#!topic/yardoc/Y700TGceoII)

# Completed Tasks

### DIFF
--- a/lib/yard/code_objects/extra_file_object.rb
+++ b/lib/yard/code_objects/extra_file_object.rb
@@ -80,7 +80,8 @@ module YARD::CodeObjects
     # @param [String] data the file contents
     def parse_contents(data)
       retried = false
-      cut_index = 0
+      cut_index_begin = 0
+      cut_index_end = 0
       data = translate(data)
       data = data.split("\n")
       data.each_with_index do |line, index|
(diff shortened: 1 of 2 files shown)


lsegal:
There are definitely concerns about breaking changes here. I realize that you show a non-breaking option, but I'm not 100% sure that dc5788a5c386997d72625c0579f4df2b17adc8ed is really the only breaking change. It seems to me that all of these parser changes would break examples where, for example, a user was attempting to render an HTML document, i.e.:

```
<!--
# a comment
-->
<html>
<head>
...
```

This would potentially incorrectly render the comment in the document, depending on its contents.

This is quite an edge case, I realize, but it should be clear that this is technically a breaking change.

I actually think that the "2 newlines" option is actually the least breaking, since newlines at the top of the file are undefined behavior. Not sure if that specific heuristic is useful to your use case though.

skalee:
Rebased for maintenance. Nothing has changed in commit contents.

lsegal:
> [!NOTE]

PR #1205, opened 2018-11-21
Ignore all magic comments at the start of file

gmalette:
# Description

YARD previously recognized three magic comments:
- shebang
- `encoding`
- `frozen_string_literal`

Standard Ruby also comes with another magic comment: `warn_ident`.

The use of magic comments is not limited to what comes built-in with
Ruby, however; some tooling uses magic comments as configuration.

It is desirable for YARD to ignore all magic comments at the beginning
of the file, in order to not clobber module documentation with magic
comments.

# Completed Tasks

### DIFF
--- a/lib/yard/parser/ruby/ruby_parser.rb
+++ b/lib/yard/parser/ruby/ruby_parser.rb
@@ -532,6 +532,8 @@ module YARD
             elsif comment =~ SourceParser::FROZEN_STRING_LINE
               @frozen_string_line = comment
               not_comment = true
+            elsif comment =~ SourceParser::MAGIC_COMMENT
+              not_comment = true
             end
           end
 
--- a/lib/yard/parser/source_parser.rb
+++ b/lib/yard/parser/source_parser.rb
@@ -65,6 +65,7 @@ module YARD
       SHEBANG_LINE  = /\A\s*#!\S+/
       ENCODING_LINE = %r{\A(?:\s*#*!.*\r?\n)?\s*(?:#+|/\*+|//+).*coding\s*[:=]{1,2}\s*([a-z\d_\-]+)}i
       FROZEN_STRING_LINE = /frozen(-|_)string(-|_)literal: true/i
+      MAGIC_COMMENT = /[\w\-_]+\s*\:.*/
 
       # The default glob of files to be parsed.
       # @since 0.9.0
--- a/spec/parser/ruby/ruby_parser_spec.rb
+++ b/spec/parser/ruby/ruby_parser_spec.rb
@@ -501,6 +501,43 @@ eof
       expect(Registry.at(:Bar).docstring).to eq "this is a comment"
     end
 
+    it "removes all magic comments at the start of the buffer" do
+      YARD.parse_string(<<~RUBY)
+        # frozen_string_literal: true
+        # frozen_string_literal : true
+        # encoding: UTF-8
+        # warn_indent: true
+        # also_magic: true
+        # More-magic: true
+        # this is a comment
+        class Foo; end
+      RUBY
+
+      expect(Registry.at(:Foo).docstring).to eq "this is a comment"
+    end
+
+    it "does not remove any magic comment past the beginning of the buffer" do
+      YARD.parse_string(<<~RUBY)
+        1
+        # frozen_string_literal: true
+        # encoding: UTF-8
+        # warn_indent: true
+        # also_magic: true
+        # More-magic: true
+        # this is a comment
+        class Foo; end
+      RUBY
+
+      expect(Registry.at(:Foo).docstring).to eq(<<~COMMENT.chomp)
+        frozen_string_literal: true
+        encoding: UTF-8
+        warn_indent: true
+        also_magic: true
+        More-magic: true
+        this is a comment
+      COMMENT
+    end
+
     it "handles compile errors" do
       expect { stmt(":~$ Do not clobber") }.to raise_error(Parser::ParserSyntaxError)
     end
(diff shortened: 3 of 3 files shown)


gmalette:
More context about this: YARD would also allow us to use magic comments and omit them from class documentation by using two line returns (or so I hear) without considering it a redefinition. However, Rubocop disallows this, considering it extra blank lines. 

We thus have to make a choice. Either Rubocop is aware of YARD and of magic comments, or YARD is aware of magic comments. I made the choice to teach YARD about magic comments because the coupling factor is lower.

lsegal:
Sorry for the late response.

I think the concern here is that the regex would be too liberal and clobber potentially valid docstrings. Even a seemingly innocuous first-line comment as follows would be silently ignored if placed at the top of a file:

```ruby
# Net::HTTP handles all HTTP networking requests for an application
class Net::HTTP; end
```

Even if special consideration was made for "::", proper names that included ':' would be problematic, like any XML node containing a colon or other library names that happened to include the character.

I would much rather be explicit about the magic comments being supported-- it seems like we only need to add an extra match for the new `warn_indent` magic comment.

As a sidenote, it seems like one improvement we can make here is to support `# -*- ... -*-` syntax, which is apparently also legal: https://idiosyncratic-ruby.com/58-magic-instructions.html

gmalette:
> I would much rather be explicit about the magic comments being supported-- it seems like we only need to add an extra match for the new `warn_indent` magic comment.

These would allow for Ruby's magic comments, but not application-defined magic comments. Perhaps we could inject those as a setting or something?

lsegal:
Can you elaborate on application defined magic comments? I'm not really aware of many besides rubocop, which really isn't a top-of-file problem as much as an everywhere-problem (so ignoring it solely from the top of a file would be misleading behavioral support).

gmalette:
Sure!
For example, I've been using comments at the top of file that say what sub-suite a test is part of. We rank them by speed in ms making it easy to run only tests that complete in say 1ms, 10ms or 100ms.

```
# slow: 10
```

Another example is Stripe's [Sorbet](https://sorbet.run/) which uses `# typed: true` to indicate that a file should be analyzed by the static type system.

lsegal:
The easiest way to work around this with YARD syntax would be to just define it as a tag via `@slow`, which also allows you to introspect those definitions in your docs later. Or, alternatively, just use 2 newlines to separate the comment from the object:

```ruby
# slow: 10


class Foo; end
```

I don't really see any feasible way to handle this generically if it's really going to parse any `X: Y` value. There's no good way to differentiate language like `typed: true` from other prose (like a mid-sentence phrase that split to a newline). This is kind of why YARD uses meta tag syntax for clear parseability. I would recommend using YARD syntax to disambiguate if you can avoid it.

It's reasonable to special case Ruby-specific syntaxes because they are part of the language and are specific sets of syntax that mean something to every running Ruby program, so are not going to be misused. Opening this up to app defined behavior is a lot of extra complexity that can easily be avoided with a number of usable workarounds, and when the worst case behavior is just an extra line in your docstring, I'm not sure that parsing complexity is worth it.

connorshea:
Would you support an option that lets the user define magic comments to ignore via a CLI flag? e.g. if I wanted YARD to ignore `# typed: true` I could add `typed` to the list of ignored magic comments.

lsegal:
@connorshea I find that to be a very specialized option / extra complexity baked into YARD given the very simple workaround and very specific use-case. If the situation is such that Sorbet (or other lib) wants to introduce a new idiom to use `# typed: true` in file headers, then it seems like a localized problem that can and should be handled by a YARD extension/plugin (i.e. a `yard-sorbet`) to ignore that comment for those types of codebases. This plus #1261 leads me to believe that such a specialized plugin could benefit users from more than just the removal of magic comments.

In any case, writing such a specialized extension for YARD can be done with a fairly small amount of code. Here's a first-pass at a generalized "ignore these magic comments" type extension:

```ruby
# place me in yardext/ignore_magic_comments.rb
module IgnoreMagicComments
  module DocstringParserAdditions
    def parse(content, object = nil, handler = nil)
      content = remove_magic_comment(content, object)
      super
    end

    private

    def remove_magic_comment(content, object)
      return content if object.nil? || object.files.size == 0

      # only strip comments if they are at the top of a file
      # we determine this by finding the object's source location and counting
      # back to the position in the file at the top of the docstring
      lineno = object.files.last.last
      at_first_line = lineno - (content.count("\n") + 1) <= 1
      return content unless at_first_line

      # parse first lines removing any magic comments we've defined
      # in the MAGIC_COMMENT_NAMES environment variable
      in_magic = true
      content.split("\n").map do |line|
        return line unless in_magic

        match = /\A(?<name>\S+):\s/.match(line)
        if match && all_magic_names.include?(match[:name])
          line = nil
        else
          in_magic = false
        end

        line
      end.compact.join("\n")
    end

    def all_magic_names
      @all_magic_names ||= (ENV['MAGIC_COMMENT_NAMES'] || '').split(',').map(&:downcase)
    end
  end

  YARD::DocstringParser.prepend DocstringParserAdditions
end
```

Then use this with:

```sh
$ MAGIC_COMMENT_NAMES=typed yard -e yardext/ignore_magic_comments.rb 
```

You can, of course, simplify by just having `all_magic_names` return a hardcoded Array for your use case. Flipping this around, this code can be turned into a plugin and shared with others. Frankly there are some good crossover opportunities with useful integrations from #1261 that Sorbet users would probably find useful if someone wants to pick this work up.

gmalette:
Seems like the right place to do this would be a plug-in, closing the issue

PR #1212, opened 2018-12-27
Generate documentation for File singleton methods

castwide:
# Description

When generating documentation from the Ruby source, some File methods do not get parsed into the yardoc.

## Steps to reproduce

This is the minimal reproduction for the issue. I've done my best to remove
all extraneous code and unique environment state on my machine before providing
these steps:

1. Run the following command in the Ruby source: `yardoc *.c`
2. Run `yri File.exist?`

## Actual Output

No documentation for `File.exist?'

## Expected Output

The documentation from the source code.

## Solution

In the Ruby core, `file.c` uses `define_filetest_function()` to define certain File singleton methods, including `File.exist?` YARD's C parser doesn't find them.

This PR modifies `YARD::Handlers::C::MethodHandler` to process `define_filetest_function`.

# Completed Tasks

* [X] I wrote tests and ran `bundle exec rake` locally (if code is attached to PR).

### DIFF
--- a/lib/yard/handlers/c/method_handler.rb
+++ b/lib/yard/handlers/c/method_handler.rb
@@ -15,8 +15,13 @@ class YARD::Handlers::C::MethodHandler < YARD::Handlers::C::Base
                 \s*"([^"]+)",
                 \s*(?:RUBY_METHOD_FUNC\(|VALUEFUNC\(|\(\w+\))?(\w+)\)?,
                 \s*(-?\w+)\s*\)/xm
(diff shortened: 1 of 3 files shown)


castwide:
I'm currently using a [plugin implementation](https://github.com/castwide/solargraph/blob/master/lib/yard-coregen.rb) of this feature. Would you prefer to keep it external? If so, we can close this PR and I can publish the plugin as a gem. That might be more practical in the long run, since there might be other exceptional cases that are specific to the Ruby core.

PR #1220, opened 2019-01-29
Fix multiline method signatures

segiddins:
# Description

Previously, the `signature` for a `MethodObject` would be cut-off when the parameters spanned multiple lines. Now, the signature will always be complete and fit on a single line.

# Completed Tasks

### DIFF
--- a/lib/yard/handlers/ruby/dsl_handler_methods.rb
+++ b/lib/yard/handlers/ruby/dsl_handler_methods.rb
@@ -41,8 +41,9 @@ module YARD
           end
 
           object = MethodObject.new(namespace, method_name, scope)
+          object = register(object)
           object.signature = method_signature
-          register(object)
+          object
         end
 
         def register_docstring(object, docstring = @docstring, stmt = statement)
--- a/lib/yard/handlers/ruby/method_handler.rb
+++ b/lib/yard/handlers/ruby/method_handler.rb
@@ -110,7 +110,17 @@ class YARD::Handlers::Ruby::MethodHandler < YARD::Handlers::Ruby::Base
   def method_signature
     method_name = statement.method_name(true)
     if statement.parameters.any? {|e| e }
-      "def #{method_name}(#{statement.parameters.source})"
+      formatted_args = format_args.map do |name, default|
+        next name unless default
+
+        if name.end_with?(':')
+          [name, default].join(' ')
+        else
+          [name, default].join(' = ')
+        end
+      end.join(', ')
+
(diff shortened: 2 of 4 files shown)


lsegal:
@segiddins can you set up conditional test cases for `RUBY_VERSION >= '2.1'` and the opposite case? alternatively you can just test with a default, since that's not really what we're exercising here.

segiddins:
Fair enough! changed all fo them to have default args

lsegal:
@segiddins were your changes pushed?

segiddins:
:facepalm:

lib/yard/handlers/ruby/dsl_handler_methods.rb
```diff
@@ -41,8 +41,9 @@ def handle_comments
           end
 
           object = MethodObject.new(namespace, method_name, scope)
+          object = register(object)
```
lsegal: nit pick here but you can probably just call `object = register MethodObject.new(...)`

lib/yard/handlers/ruby/method_handler.rb
```diff
@@ -19,10 +19,10 @@ class YARD::Handlers::Ruby::MethodHandler < YARD::Handlers::Ruby::Base
 
     nobj = P(namespace, nobj.value) while nobj.type == :constant
     obj = register MethodObject.new(nobj, meth, mscope) do |o|
-      o.signature = method_signature
       o.explicit = true
       o.parameters = args
     end
+    obj.signature = method_signature
```
lsegal: It seems like this should be fixed in `MethodObject.new` directly-- I think it's reasonable to expect that setting `o.signature` inside the yielded block would override the default signature, so this is probably an ordering bug inside of #initialize rather than here.

lib/yard/handlers/ruby/dsl_handler_methods.rb
segiddins: see the commit message for https://github.com/lsegal/yard/pull/1220/commits/94ecc1de793d1046ca5c58cb463aaf6c28f25dc0, but the gist is im pretty sure the stomping was happening in `register`, when we assign the source

segiddins: ping on this response?

segiddins:
I _think_ I responded to both of your concerns, it kind of boils down to the fast that `register` is what was doing the stomping when assigning the source, so I couldn't just make the fix happen in `MethodObject.new`

PR #1226, opened 2019-02-25
Fix double return tag.

thomthom:
# Description

In our C/C++ code base where we implement Ruby functions we some times have explicit `@return [Boolean]` tags for predicate methods.

YARD is adding an extra `@return [Boolean]` regardless if there already is one.

```cpp
/**
 * This method is used to determine if a bounding box contains a specific
 * Point3d or BoundingBox object.
 *
 * @example
 *   boundingbox = Geom::BoundingBox.new
 *   boundingbox.add([100, 200, -400], [200, 400, 100])
 *   # This will return false.
 *   boundingbox.contains?([300, 100, 400])
 *   # This will return true.
 *   boundingbox.contains?([150, 300, -200])
 *
 * @overload contains?(point_or_bb)
 *
 *   @param [Geom::Point3d, Geom::BoundingBox] point_or_bb
 *
 * @return [Boolean]
 *
 * @version SketchUp 6.0
 */
static VALUE _wrap_contains(VALUE self, VALUE arg) {
  // ...
}
```

image

See output: http://ruby.sketchup.com/Geom/BoundingBox.html#contains%3F-instance_method

# Completed Tasks

### DIFF
--- a/lib/yard/handlers/c/handler_methods.rb
+++ b/lib/yard/handlers/c/handler_methods.rb
@@ -5,6 +5,7 @@ module YARD
       module HandlerMethods
         include Parser::C
         include CodeObjects
+        include Common::MethodHandler
 
         def handle_class(var_name, class_name, parent, in_module = nil)
           parent = nil if parent == "0"
@@ -67,7 +68,7 @@ module YARD
             register_visibility(obj, visibility)
             find_method_body(obj, func_name)
             obj.explicit = true
-            obj.add_tag(Tags::Tag.new(:return, '', 'Boolean')) if name =~ /\?$/
+            add_predicate_return_tag(obj) if name =~ /\?$/
           end
         end
 
--- a/lib/yard/handlers/common/method_handler.rb
+++ b/lib/yard/handlers/common/method_handler.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module YARD::Handlers
+  module Common
+    # Shared functionality between Ruby and C method handlers.
+    module MethodHandler
+      # @param [MethodObject] obj
+      def add_predicate_return_tag(obj)
+        if obj.tag(:return) && (obj.tag(:return).types || []).empty?
+          obj.tag(:return).types = ['Boolean']
+        elsif obj.tag(:return).nil?
+          unless obj.tags(:overload).any? {|overload| overload.tag(:return) }
+            obj.add_tag(YARD::Tags::Tag.new(:return, "", "Boolean"))
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/handlers/c/method_handler_spec.rb
+++ b/spec/handlers/c/method_handler_spec.rb
@@ -213,6 +213,7 @@ RSpec.describe YARD::Handlers::C::MethodHandler do
     foo = Registry.at('Foo#foo?')
     expect(foo.docstring).to eq 'DOCSTRING'
     expect(foo.tag(:return).types).to eq ['Boolean']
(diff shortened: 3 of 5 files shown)


spec/handlers/c/method_handler_spec.rb
```diff
@@ -213,6 +213,7 @@
     foo = Registry.at('Foo#foo?')
     expect(foo.docstring).to eq 'DOCSTRING'
     expect(foo.tag(:return).types).to eq ['Boolean']
+    expect(foo.tags(:return).size).to eq 1
```
thomthom: Without the fix to `handler_methods.rb` this test still passed. And I don't quite understand why. I would have expected this to yield 2 and fail when YARD added an extra `@return` tag.

spec/handlers/c/method_handler_spec.rb
```diff
@@ -226,6 +227,7 @@
     eof
     foo = Registry.at('Foo#foo?')
     expect(foo.tag(:return).types).to eq ['String']
+    expect(foo.tags(:return).size).to eq 1
```
thomthom: This is the test that catches the bug. I can dig deeper into why the other test didn't raise any errors before the fix.

spec/handlers/c/method_handler_spec.rb
lsegal: This is probably because one of the tags ends up in your overload. You probably want an implementation more normalized with the Ruby handler:

https://github.com/lsegal/yard/blob/master/lib/yard/handlers/ruby/method_handler.rb#L41-L49

lib/yard/handlers/c/handler_methods.rb
```diff
@@ -67,7 +67,9 @@ def handle_method(scope, var_name, name, func_name, _source_file = nil)
             register_visibility(obj, visibility)
             find_method_body(obj, func_name)
             obj.explicit = true
-            obj.add_tag(Tags::Tag.new(:return, '', 'Boolean')) if name =~ /\?$/
+            if name =~ /\?$/ && obj.tags(:return).empty?
```
lsegal: This implementation should probably be normalized to handle overloads as per the Ruby handler:

https://github.com/lsegal/yard/blob/master/lib/yard/handlers/ruby/method_handler.rb#L41-L49

spec/handlers/c/method_handler_spec.rb
lsegal: There are no extra tags on the method defined in this specific test, which would explain why the result hasn't changed.

lib/yard/handlers/c/handler_methods.rb
thomthom: Is there any existing patterns to sharing logic between Ruby and C handlers?
Or should I just copy+paste?

thomthom:
I took another stab at this. Added a `Common` module under `handlers` for share logic.

I'll squash it if you are ok with these changes.

lib/yard/handlers/common/method_handler.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+# Shared functionality between Ruby and C method handlers.
```
lsegal: It looks like this will add a docstring to the `YARD::Handlers` module, not MethodHandler.

lsegal: You will also want at least one newline after the frozen directive

thomthom:
I corrected the docstring and rebased to latest YARD master branch.

lsegal:
Great thanks for your work on this!

PR #1233, opened 2019-03-30
resolves #1231 sort README files

### LINKED ISSUE #1231: Prefer the README file that has the basename README
I'm facing a situation where the translated version of the README (e.g., README-jp.adoc) in a gem is being selected over the primary (English) one (README.adoc). (See https://www.rubydoc.info/gems/asciidoctor/2.0.0). This happens because of how the the README files are globbed.

The README files in the gem are as follows:

```
README.adoc
README-de.adoc
README-fr.adoc
README-jp.adoc
README-zh_CN.adoc
```

The glob used by YARD is `README*`. For some reason, that results in README-jp.adoc being selected as the default. (It appears the result of the glob isn't sorted at all).

I think the file with the basename README (e.g., README.adoc) should be preferred over files that are hyphenated. I think most people would agree that this is the master README file in all situations.

(I'm aware I can include a .yardopts file to select a specific README. However, I can't make that change for gems that have already been released).

mojavelinux:
# Description

The main reason for this change is to choose the README that users most expect to be chosen. Currently, there isn't really any rhyme or reason to which README is selected because the files aren't explicitly sorted. With this change, README will be selected over README.md, and README.md will be selected over README-zh_CN.md (or README.zh_CN.md).

The changes included are:

- sort README files, first by file extension, then by basename
- prefer README basename over all other files
- prefer file without extension over file with extension

# Completed Tasks

### DIFF
--- a/lib/yard/cli/yardoc.rb
+++ b/lib/yard/cli/yardoc.rb
@@ -294,8 +294,15 @@ module YARD
         # Last minute modifications
         self.files = Parser::SourceParser::DEFAULT_PATH_GLOB if files.empty?
         files.delete_if {|x| x =~ /\A\s*\Z/ } # remove empty ones
-        readme = Dir.glob('README{,*[^~]}').first
-        readme ||= Dir.glob(files.first).first if options.onefile
+        readmes = Dir.glob('README{,*[^~]}')
+        if readmes.empty?
+          readme = Dir.glob(files.first).first if options.onefile && !files.empty?
+        else
+          readme = readmes.
+            sort {|a, b| File.extname(a) <=> File.extname(b) }.
+            sort {|a, b| a.slice(0, a.rindex('.') || a.length) <=> b.slice(0, b.rindex('.') || b.length) }.
+            first
+        end
         options.readme ||= CodeObjects::ExtraFileObject.new(readme) if readme
         options.files.unshift(options.readme).uniq! if options.readme
 
--- a/Gemfile
+++ b/Gemfile
@@ -13,8 +13,7 @@ group :development do
 end
 
 group :asciidoc do
-  # Asciidoctor 2.0 drops support for Ruby < 2.3.
-  gem 'asciidoctor', RUBY_VERSION < '2.3' ? '< 2' : '>= 0'
+  gem 'asciidoctor'
 end
 
 group :markdown do
--- a/spec/cli/yardoc_spec.rb
+++ b/spec/cli/yardoc_spec.rb
@@ -647,11 +647,45 @@ RSpec.describe YARD::CLI::Yardoc do
       expect(@yardoc.options.readme).to eq CodeObjects::ExtraFileObject.new('lib/foo.rb', '')
     end
 
+    it "uses no readme if files is empty and no readme is specified when using --one-file" do
+      expect(Dir).to receive(:glob).with('README{,*[^~]}').and_return []
+      @yardoc.parse_arguments '--one-file', ''
+      expect(@yardoc.options.readme).to be_nil
+    end
+
     it "uses readme it exists when using --one-file" do
       expect(Dir).to receive(:glob).with('README{,*[^~]}').and_return ['README']
       expect(File).to receive(:read).with('README').and_return('')
       @yardoc.parse_arguments(*%w(--one-file lib/*.rb))
       expect(@yardoc.options.readme).to eq CodeObjects::ExtraFileObject.new('README', '')
+    end
+
+    it "selects readme with no file extension over readme with file extension" do
+      expect(Dir).to receive(:glob).with('README{,*[^~]}').and_return ['README.md', 'README']
+      expect(File).to receive(:read).with('README').and_return('')
+      @yardoc.parse_arguments
+      expect(@yardoc.options.readme).to eq CodeObjects::ExtraFileObject.new('README', '')
+    end
+
+    it "selects readme with no suffix over readme with hyphenated suffix" do
+      expect(Dir).to receive(:glob).with('README{,*[^~]}').and_return ['README-fr.md', 'README.md', 'README-de.md']
+      expect(File).to receive(:read).with('README.md').and_return('')
+      @yardoc.parse_arguments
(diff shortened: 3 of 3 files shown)


skalee:
Tests should succeed after merging #1234 in.

lib/yard/cli/yardoc.rb
```diff
@@ -294,7 +295,16 @@ def parse_arguments(*args)
         # Last minute modifications
         self.files = Parser::SourceParser::DEFAULT_PATH_GLOB if files.empty?
         files.delete_if {|x| x =~ /\A\s*\Z/ } # remove empty ones
-        readme = Dir.glob('README{,*[^~]}').first
+        readmes = Dir.glob('README{,*[^~]}')
+        if readmes.empty?
+          readme = Dir.glob(files.first).first if options.onefile
```
skalee: If I see correctly, this branch is doubled by line no. 308.

lsegal: This is correct. Seems like this can be golfed down to just:

```ruby
readme = Dir.glob('README{,*[^~]}').map(....)....first
```

mojavelinux: I didn't make that change because it goes outside the scope of the original issue (as it deals with a separate optimization / case). If you'd like me to incorporate that change into this patch, I'd be happy to update it.

lib/yard/cli/yardoc.rb
```diff
@@ -294,7 +295,16 @@ def parse_arguments(*args)
...
+        else
+          readme = readmes.map {|file| Pathname file }.
```
lsegal: Please use `File.extname` instead of pulling in Pathname, since it is [significantly slower](https://github.com/lsegal/yard/blob/master/benchmarks/pathname_vs_string.rb).

mojavelinux: That's an outdated benchmark. Pathname is now C-accelerated, so nothing is faster than Pathname operations.

mojavelinux: (And that's compounded by the fact that we do many more operations than just a single extname call).

mojavelinux: I just did a benchmark to verify my statement. As it turns out, File.extname is still sightly faster assuming you already have a Pathname.

```
        File.extname:  2751417.2 i/s
    Pathname#extname:  2423775.1 i/s - 1.14x  slower
```

But that's not the only operation we are doing here. Let me see how the whole chain compares.

mojavelinux: Though I still hold to my original statement that Pathname is no longer as slow as it once was because it's now C-accelerated. We have to be careful not to carry the fallacies about Pathname forward.

mojavelinux: Despite the improvements to Pathname, it looks like not using Pathname is still faster. If we really want it to be as accelerated as possible, we wouldn't even use File.

```rb
readmes.
  sort {|a, b| a.slice(a.rindex('.') || a.length, a.length) <=> b.slice(b.rindex('.') || b.length, b.length) }.
  sort {|a, b| a.slice(0, a.rindex('.') || a.length) <=> b.slice(0, b.rindex('.') || b.length) }[0]
```

lsegal: The benchmark file in the repo is still reporting similar timings in Ruby 2.5.x:

```
λ ruby pathname_vs_string.rb
join:
Rehearsal -------------------------------------------------------
pathname              0.015000   0.000000   0.015000 (  0.021441)
string                0.000000   0.000000   0.000000 (  0.000626)
pathname-sameobject   0.032000   0.000000   0.032000 (  0.023407)
string-sameobject     0.000000   0.000000   0.000000 (  0.000630)
---------------------------------------------- total: 0.047000sec

                          user     system      total        real
pathname              0.015000   0.000000   0.015000 (  0.020978)
string                0.000000   0.000000   0.000000 (  0.000567)
pathname-sameobject   0.031000   0.000000   0.031000 (  0.020176)
string-sameobject     0.000000   0.000000   0.000000 (  0.000591)


cleanpath:
Rehearsal --------------------------------------------
pathname   0.000000   0.000000   0.000000 (  0.005434)
string     0.000000   0.000000   0.000000 (  0.001758)
----------------------------------- total: 0.000000sec

               user     system      total        real
pathname   0.000000   0.000000   0.000000 (  0.005791)
string     0.000000   0.000000   0.000000 (  0.001719)
```

The cost is in unnecessary allocations, not the C acceleration. The problem will get worse as you have more Pathname objects to create. There's still a valid concern here of bringing in an extra lib for a single Ruby method call that's well represented in the core functions. 

I don't think using slice is necessary. `File.extname` is a reasonable balance.

mojavelinux: :+1:

mojavelinux: I'll update the PR accordingly.

lib/yard/cli/yardoc.rb
mojavelinux: Nevermind, I get it now. You're right.

mojavelinux: I'll add a test.

mojavelinux:
PR updated as requested.

mojavelinux:
Build is green.

lib/yard/cli/yardoc.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require 'digest/sha1'
 require 'fileutils'
+require 'pathname'
```
lsegal: nit but we can remove this require now.

mojavelinux:
Doh! Of course. I'll update.

lib/yard/cli/yardoc.rb
mojavelinux: Done.

lsegal:
Thanks for working on this!

mojavelinux:
You're welcome. Glad to help!

lsegal:
Seems as though there are some failures on Windows due to this logic, probably conflicting with Dir.glob on other platforms. I'm looking into this now.

mojavelinux:
That's very unexpected. I'll be interested to hear what you find.

PR #1235, opened 2019-03-31
switch to official Asciidoctor API from the deprecated one

mojavelinux:
# Description

Asciidoctor.render -> Asciidoctor.convert

# Completed Tasks


mojavelinux:
PR updated with an additional test to extend coverage.

mojavelinux:
Correct. Both method names have been mapped since 1.5.0 (which was the 1.0 release). There should be no one using Asciidoctor < 1.5.0 at this point, so this is just a matter of dropping the use of the deprecated method so we can plan to remove it from Asciidoctor (probably in 3 or 4).

Issue #1243, opened 2019-05-09
YARD fails to recognize empty symbol (:"") on Ruby 2.6.3, which causes build failures


skalee:
Something has changed in Ruby 2.6.3. In the result, two tests fail.

## Steps to reproduce

1. Switch to Ruby 2.6.3.
2. Run `bundle exec rspec` or `bundle exec rake`.

## Actual Output

<details><summary>RSpec Failures</summary>

```
  1) YARD::Handlers::Ruby::ConstantHandler documents block for Struct.new if present
     Failure/Error: expect(obj.constants[1].value).to eq ':""'

       expected: ":\"\""
            got: ":\"\\n\""

       (compared using ==)
     # ./spec/handlers/constant_handler_spec.rb:39:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:127:in `block (2 levels) in <top (required)>'

  2) YARD::Templates::Helpers::MethodHelper#format_constant displays correctly constant values which are quoted symbols
     Failure/Error: expect(format_constant(foo)).to eq ":&quot;&quot;"

       expected: ":&quot;&quot;"
            got: ":&quot;\\n&quot;"

       (compared using ==)
     # ./spec/templates/helpers/method_helper_spec.rb:102:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:127:in `block (2 levels) in <top (required)>'
```

</details>

## Expected Output

All tests should pass.

## Possible causes

According to release notes, not much has changed:

> This release adds support for New Japanese Era “令和” (Reiwa). It updates the Unicode version to 12.1 beta and updates the date library.

Full changelog:

- https://github.com/ruby/ruby/compare/v2_6_2...v2_6_3

Main suspect:

- https://github.com/ruby/ruby/commit/8bf06cb9752273f8fa1c9e176bb5ba85381f22a5 which is supposed to fix https://bugs.ruby-lang.org/issues/15670

Secondary suspect:

- https://github.com/ruby/ruby/commit/69ec39363eb64fd492af9fee3a7d0188f11542df

----

I have read the [Contributing Guide][contrib].

[contrib]: https://github.com/lsegal/yard/blob/master/CONTRIBUTING.md

skalee:
More details:

In ./spec/handlers/constant_handler_spec.rb:38, source range of `s(:string_content, )` is `417..417`, whereas it should be `416..415`.

lsegal:
This was fixed in #1256. Thanks for reporting!

PR #1244, opened 2019-05-24
Update README section on how to use a Rake task (NameErrors)

skalee:
Currently, README is incorrect about how to setup YARD's Rake task. And Rake task itself is also broken.

# Description

There is an example in README which explains how to add a Rake task. However, following it results with a `NameError` due to uninitialized constant `YARD`. This is not surprising, because Rake task definition should be loaded first, which is missing in that example. Unfortunately, adding `require 'yard/rake/yardoc_task'` does not help much — another `NameError` is raised, because the Rake task was unable to load YARD.

Changes included in this pull request:

1. Ensure that YARD is loaded when a Rake task is invoked.
2. Update README with a corrected example.

# Completed Tasks

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,8 @@ The second most obvious is to generate docs via a Rake task. You can do this by
 adding the following to your `Rakefile`:
 
 ```ruby
+require 'yard'
+
 YARD::Rake::YardocTask.new do |t|
  t.files   = ['lib/**/*.rb', OTHER_PATHS]   # optional
  t.options = ['--any', '--extra', '--opts'] # optional


skalee:
Test failures for Ruby 2.6.3 are unrelated to this pull request, but caused by #1243.

lsegal:
You should not require the `yardoc_task` directly, instead documentation should be updated to suggest requiring 'yard'.

README.md
```diff
@@ -226,6 +226,8 @@ The second most obvious is to generate docs via a Rake task. You can do this by
 adding the following to your `Rakefile`:
 
 ```ruby
+require 'yard/rake/yardoc_task'
```
lsegal: Switching this to `yard` fixes the issue around lazy loading below.

```suggestion
require 'yard'
```

skalee:
I am okay with either solution. Updated, please merge.

README.md
skalee: Ok.

lsegal:
Thanks!

PR #1260, opened 2019-07-04
Improve performance by 10%

Confusion:
Inspired by a blog post [1] on the 'allocation_stats' gem I looked into
yard's memory use (using the 'memory_profiler' gem) and found some hotspots
that the attached patch removes. It reduces memory allocations by ~25%,
while increasing performance by ~10%, measured on the test_doc Rake task of
Yard itself.

[1] https://blog.presidentbeef.com/blog/2018/11/28/finding-ruby-performance-hotspots-via-allocation-stats/

# Description

# Completed Tasks

### DIFF
--- a/lib/yard/registry_resolver.rb
+++ b/lib/yard/registry_resolver.rb
@@ -62,7 +62,7 @@ module YARD
       end
       orignamespace = namespace
 
-      if path =~ /\A#{default_separator}/
+      if path =~ starts_with_default_separator_match
         path = $'
         namespace = @registry.root
         orignamespace = @registry.root
@@ -102,12 +102,12 @@ module YARD
       result = namespace.root? && validate(@registry.at(path), type)
       return result if result
 
-      if path =~ /\A(#{separators_match})/
+      if path =~ starts_with_separator_match
         return validate(@registry.at(namespace.path + path), type)
       end
 
       separators.each do |sep|
-        result = validate(@registry.at(namespace.path + sep + path), type)
+        result = validate(@registry.at("#{namespace.path}#{sep}#{path}"), type)
         return result if result
       end
 
@@ -122,12 +122,12 @@ module YARD
       last_sep = nil
       pos = 0
 
-      if path =~ /\A(#{separators_match})/
+      if path =~ starts_with_separator_match
         last_sep = $1
         path = $'
       end
 
-      path.scan(/(.+?)(#{separators_match}|$)/).each do |part, sep|
+      path.scan(split_on_separators_match).each do |part, sep|
         cur_obj = nil
         pos += "#{part}#{sep}".length
         parsed_end = pos == path.length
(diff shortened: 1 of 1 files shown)


lib/yard/code_objects/namespace_mapper.rb
```diff
@@ -68,6 +73,14 @@ def separators_match
         NamespaceMapper.map_match
       end
 
+      def starts_with_separator_regex
```
lsegal: These two methods have a possible caching bug-- if `NamespaceMapper.invalidate` is called, these will not be invalidated along with `map_match`.

Confusion:
I renamed the methods, moved them to RegistryResolver as private methods (but documented them anyway) and hooked into `invalidate` and `default_seperator` to invalidate the memoizations when necessary.

This of course creates additional coupling between the RegistryResolver and the NamespaceMapper through the 'super' calls (which at least will break on renames and changes in number of parameters, so chance of silently losing functionality seems slim), but I don't think there is any way to avoid that.

Performance gain is still ~10% (actually it seems more like 15-20%, but I do not sufficiently trust my measurement methodology to claim that much)

lib/yard/registry_resolver.rb
```diff
@@ -185,5 +185,35 @@ def collect_namespaces(object)
 
       nss
     end
+
+    # @see NamespaceMapper#invalidate
+    def invalidate
```
lsegal: Unfortunately this won't trigger as `Namespacemapper.invalidate` is a module function. 

The good news is I think you can/should override `#clear_separators` instead, since NamespaceMapper's singleton stuff internal and only expected to be used _through_ / exposed by those instance methods.

```suggestion
    def clear_separators
```

(Need to change the docstring too)

lib/yard/registry_resolver.rb
```diff
@@ -185,5 +185,35 @@ def collect_namespaces(object)
 
       nss
     end
+
```
lsegal: You should probably mark all these methods below as private since they are not part of the public API.

```suggestion

    private

```

lib/yard/registry_resolver.rb
Confusion: :+1: 

I now realize my previous comment wasn't true though: if `clear_separators` is renamed, this won't fail: it just silently won't be called. Do you want me to add a spec for that?

lib/yard/registry_resolver.rb
Confusion: There's a `private` marker on line 93 that covers these methods, so that's already taken care of?

lsegal: You're right about that, I missed it! You can ignore this one.

lib/yard/registry_resolver.rb
lsegal: Sure a spec would help.

Confusion: I just noticed that `invalidate` is also called via `register_separator`, so `register_separator` won't result in the memoizations being cleared, which seems necessary. So we need to extend `invalidate` after all? 

What we could do is add a line
```
invalidate if respond_to?(:invalidate)
```
to `register_separator`, after `NamespaceMapper.invalidate` so users of NamespaceMapper can define an instance method :invalidate for additional resets. What do you think about that?

Confusion: I opted to extend both #register_separator and #clear_separators

Issue #1259, opened 2019-07-01
Consider updating HTML so can pass W3C validator


esotericpig:
The [W3C validator](https://validator.w3.org) complains about 2 issues regarding the generated HTML:

1. Script tags should not define `charset="utf-8"`
2. A language tag should be added to the HTML tag.

For Number 1, [headers.erb](https://github.com/lsegal/yard/blob/master/templates/default/layout/html/headers.erb) would need to be modified to remove `charset=` in Script tags.

Number 2 is not a concern I think, but would be nice to specify as a command-line option, like `yardoc lang=en`.

## Steps to reproduce

1. Run the following command: `yard doc`
2. Copy produced HTML source from "index.html" (or any others)
3. Go to https://validator.w3.org/#validate_by_input
4. Paste in HTML source from Step 2
5. Click "Check"

## Actual Output

`Error: The charset attribute on the link element is obsolete. Use an HTTP Content-Type header on the linked resource instead.`

`Warning: This document appears to be written in English. Consider adding lang="en" (or variant) to the html start tag.`

## Expected Output

No errors/warnings.

lsegal:
Thanks for bringing this up. This would be an easy PR to take on if you're interested!

esotericpig:
Thanks for merging my pull request.

HTML lang is just a warning, so probably not needed.

PR #1263, opened 2019-07-08
Remove charset attr from link and script tags (if no src attr)

esotericpig:
# Description

Issue #1259

According to the new HTML 5 standards, link tags should never use the charset attribute (charset should be specified in the CSS files) and script tags should only specify it if there is a src attribute.

In the [W3C validator](https://validator.w3.org/), it outputs them as errors, instead of warnings, for some reason.

https://www.w3schools.com/tags/att_link_charset.asp
https://www.w3schools.com/tags/att_script_charset.asp

# Completed Tasks


lsegal:
Thanks for doing this!

PR #1275, opened 2019-08-29
Map Ruby's C variable error names to their Ruby names. Fixes #1270

### LINKED ISSUE #1270: Inheriting from core error classes in C output incorrect name
It appear that YARD isn't resolving the names of core error classes when the classes are defined in C source.

I'm guessing there needs to be a a lookup table to map the built in error classes to their real names.

Not sure where that's best done though. (Ideas?)

I see that the error class ends up as a proxy:

image

Is it perhaps possible to pre-populate YARD with definitions of the C error classes?

I see the classes are defined in `error.c` in Ruby's source. https://github.com/ruby/ruby/blob/930e637c0e2ffd6fcb46205d473b6f4c05607471/error.c#L2478
Maybe a tool to scrape this file and periodically update the lookup table? Not that I'd expect this to update often.

## Steps to reproduce

https://github.com/thomthom/yard/commit/e064775e1f2dba0b3c6c96a5959958cf6b6d99b8

```ruby
    # spec/parser/c_parser_spec.rb

    describe "Class inherited from core error class" do
      it "resolves correct name" do
        parse <<-eof
          void
          Init_Mask(void)
          {
              rb_cMyError = rb_define_class("MyError", rb_eArgError);
          }
        eof
        klass = Registry.at('MyError')
        expect(klass.name).to eq :MyError
        expect(klass.title).to eq 'MyError'
        expect(klass.superclass.name).to eq :ArgumentError
        expect(klass.superclass.title).to eq 'ArgumentError'
      end
    end
```

## Actual Output

`rb_eArgError` ends up as `ArgError` instead of `ArgumentError`.

Can be seen in production here: http://ruby.sketchup.com/Layout/LockedEntityError.html

image

## Expected Output

[Describe what you *expected* to happen.]

## Environment details:

* OS: Windows 10 (1905)
* Ruby version (`ruby -v`): ruby 2.5.3p105 (2018-10-18 revision 65156) [x64-mingw32]
* YARD version (`yard -v`): yard 0.9.20

thomthom:
# Description

PR to address the mapping of core Ruby error names when using the Ruby C API.
https://github.com/lsegal/yard/issues/1270

I added a Rake task to scrape the master Ruby source for the names of the error classes:

```
bundle exec rake update_error_map
```

There' wasn't any existing `.rake` files in this project, so I modelled it after what RuboCop does.

# Completed Tasks


lsegal:
This took a while but thanks for the work on this!

PR #1276, opened 2019-08-30
Fix Asciidoc syntax highlight

### LINKED ISSUE #1239: YARD fails to recognize code listings in AsciiDoc documents (no links nor syntax highlight)
Consider following example output (a border was added by me for clarity):

<table><tr><td>

issue2a

</table></tr></td>

Rendering issues are apparent:

- some code samples are not in callouts
- some code samples are not highlighted
- in some code samples there are no links to classes and methods

### The problematic document

A following piece of AsciiDoc was used to generate the first section of above document:

```adoc
=== Ruby snippet (language specified)

[source,ruby]
----
two = 1 + 1
eleven = "1" + "1"
raise if Failure.problem
----
```

It translates to following HTML:

```html
<div class="sect2">
<h3 id="_ruby_snippet_language_specified">Ruby snippet (language specified)</h3>
<div class="listingblock">
<div class="content">
<pre class="highlight"><code class="language-ruby" data-lang="ruby">two = 1 + 1
eleven = "1" + "1"
raise if Failure.problem</code></pre>
</div>
</div>
</div>
```

Full document is in this Gist: https://gist.github.com/skalee/fe1a52a797f7c821ae354a59f7812fd9.

## Why it fails

The `YARD::Templates::Helpers::HtmlHelper#parse_codeblocks` method is responsible for post processing the HTML output produced by markup processor. It enables code highlighting, adds links to particular Ruby methods and classes, and finally normalizes the HTML document so that it can be styled with CSS:

https://github.com/lsegal/yard/blob/12f56cf7d58e7025085f00b9f9f2f62c24b13d93/lib/yard/templates/helpers/html_helper.rb#L624-L643

Unfortunately, the regular expression in this method fails to match code listings produced by Asciidoctor when they are annotated with name of programming language.

## Other considerations

The most obvious solution would be to make that regular expression more liberal. However:

* This regular expression is overcomplicated already.
* HTML attributes may contain other things than language names. For example, `<pre class="highlight">` in aforementioned example. Also, language name may be decorated with some additional specifiers, for example `<code class="language-ruby">` in the same example.

Therefore, some smarter detection should be discussed. I have two proposals.

### Option 1: Regular expression should be markup-processor-specific.

Method `parse_codeblocks(html)` is called from method `htmlify(text, markup = options.markup)` only. Therefore, an additional argument can be added: `parse_codeblocks(html, markup)`.

Then, a brand new regular expression can be introduced for the Asciidoctor markup processor. Other markup processors can have their own dedicated regular expressions if they need any. The current regular expression will be used as a default generic one.

### Option 2: Single regular expression with much smarter detection

A very generic regular expression can recognize different HTML attributes and prioritize them, either by value relevance (looking for a value which resembles a programming language name, like `ruby`, `language-ruby`, `code-ruby`), or by name relevance (looking for attributes with names `data-lang`, `lang`, `class` in that order).

----

Possibly the same problem as in #781, although that bug report was very unclear.

I can help implementing either, but discussion is required IMO. Personally, I am leaning towards option 1.

skalee:
# Description

This pull request improves the way in which code listing blocks in extra file objects are matched.
In the result, listings in extra file objects written in AsciiDoc are now properly enriched with YARD: the syntax is highlighted, and the identifiers are linked with their respective documentation.

Fixes #1239 (went with option 2).

# Details

The regular expression used for code blocks detection has been significantly relaxed. The updated one makes no assumptions on what HTML attributes can be set on `<pre>` or `<code>` element, allowing everything. It still requires that the listing's outermost element is `<pre>`, but this seems to be a reasonable limitation.

The programming language detection still considers the values of `class` and `lang` HTML attributes. Moreover, `data-lang` attribute, which is produced by AsciiDoctor markup processor, is also supported now.

Finally, this pull request properly ensures that code blocks produced from Ruby source files are not highlighted twice. It wasn't a problem before only because of a very lucky situation that code blocks produced by `html_markup_ruby` have a slightly different attributes, and that was preventing them from being erroneously double-highlighted.

# Completed Tasks

### DIFF
--- a/lib/yard/templates/helpers/html_helper.rb
+++ b/lib/yard/templates/helpers/html_helper.rb
@@ -65,7 +65,7 @@ module YARD
           html = html.encode(:invalid => :replace, :replace => '?')
         end
         html = resolve_links(html)
-        unless [:text, :none, :pre].include?(markup)
+        unless [:text, :none, :pre, :ruby].include?(markup)
           html = parse_codeblocks(html)
         end
         html
--- a/lib/yard/templates/helpers/html_helper.rb
+++ b/lib/yard/templates/helpers/html_helper.rb
@@ -628,11 +628,13 @@ module YARD
       # @return [String] highlighted html
       # @see #html_syntax_highlight
       def parse_codeblocks(html)
-        html.gsub(%r{<pre\s*(?:lang="(.+?)")?>(?:\s*<code\s*(?:class="(.+?)")?\s*>)?(.+?)(?:</code>\s*)?</pre>}m) do
+        html.gsub(%r{<pre((?:\s+\w+="(?:.+?)")*)\s*>(?:\s*<code((?:\s+\w+="(?:.+?)")*)\s*>)?(.+?)(?:</code>\s*)?</pre>}m) do
           string = $3
+
           # handle !!!LANG prefix to send to html_syntax_highlight_LANG
           language, = parse_lang_for_codeblock(string)
-          language ||= $1 || $2 || object.source_type
+          language ||= detect_lang_in_codeblock_attributes($1, $2)
+          language ||= object.source_type
 
           if options.highlight
             string = html_syntax_highlight(CGI.unescapeHTML(string), language)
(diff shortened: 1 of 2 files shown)


lsegal:
Sorry for the delay here. Thanks for this!

skalee:
@lsegal Any chance for release? There are some gems which will benefit from this one, e.g. https://rubydoc.info/gems/enmail.

PR #1284, opened 2019-10-30
Make sure deleted object's entries are removed from object_types

tomoasleep:
# Description

Previously, `YARD::RegistryStore#delete` removes an object entry from `@store` but does not remove its entry from `@object_types`.
Therefore, some elements of `YARD::Registry.all` may be `nil` after an object deletion such as the following code.

```console
$ pry
[1] pry(main)> require 'yard'
=> true
[2] pry(main)> YARD::Registry.register(foo = YARD::CodeObjects::ModuleObject.new(nil, :Foo))
=> #<yardoc module Foo>
[3] pry(main)> YARD::Registry.register(bar = YARD::CodeObjects::ModuleObject.new(nil, :Bar))
=> #<yardoc module Bar>
[4] pry(main)> YARD::Registry.all(:module)
=> [#<yardoc module Foo>, #<yardoc module Bar>]
[5] pry(main)> YARD::Registry.delete(foo)
=> #<yardoc module Foo>
[6] pry(main)> YARD::Registry.all(:module)
=> [nil, #<yardoc module Bar>]
```

This PR makes sure that `YARD::RegistryStore#delete` remove the deleted object's entry from `@object_types` to fix this behavior.

# Completed Tasks

### DIFF
--- a/lib/yard/registry_store.rb
+++ b/lib/yard/registry_store.rb
@@ -72,7 +72,12 @@ module YARD
     # Deletes an object at a given path
     # @param [#to_sym] key the key to delete
     # @return [void]
-    def delete(key) @store.delete(key.to_sym) end
+    def delete(key)
+      if @store[key.to_sym]
+        @object_types[@store[key.to_sym].type].delete(key.to_s)
+        @store.delete(key.to_sym)
+      end
+    end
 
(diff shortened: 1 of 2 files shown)


lsegal:
Thanks for this PR!

PR #1289, opened 2019-11-13
Store extra files in their own directory

### LINKED ISSUE #1277: Extra files with same file name
There's an issue when adding multiple extra files from different locations but having the same file name / e.g. `README.md`/. The generated documentation lists the extra files but links all to the same content.

## Steps to reproduce

This is the minimal reproduction for the issue. I've done my best to remove
all extraneous code and unique environment state on my machine before providing
these steps:

1. create 3 `README.md` files with different content - `README.md`, doc/README.md` and `tasks/README.md`
2. run `yard doc README.md,docs/README.md,tasks/README.md -o documentation/`
3. open `documentation/index.html` in the browser
4. Click on the Files link

The 3 readme files will be listed. The one in the root folder will be recognised as the main document but the other 2 links will point to the same content.

## Actual Output

```                                                                                                                        
[debug]: Parsing ["{lib,app}/**/*.rb", "ext/**/*.{c,cc,cxx,cpp,rb}"] with `ruby` parser
[debug]: Serializing to .yardoc/objects/root.dat
[debug]: Re-generating object ...
[debug]: Generating asset js/jquery.js
[debug]: Serializing to documentation/js/jquery.js
[debug]: Generating asset js/app.js
[debug]: Serializing to documentation/js/app.js
[debug]: Generating asset js/full_list.js
[debug]: Serializing to documentation/js/full_list.js
[debug]: Generating asset css/style.css
[debug]: Serializing to documentation/css/style.css
[debug]: Generating asset css/common.css
[debug]: Serializing to documentation/css/common.css
[debug]: Generating asset css/full_list.css
[debug]: Serializing to documentation/css/full_list.css
[debug]: Generating asset class_list.html
[debug]: Serializing to documentation/class_list.html
[debug]: Generating asset method_list.html
[debug]: Serializing to documentation/method_list.html
[debug]: Generating asset file_list.html
[debug]: Serializing to documentation/file_list.html
[debug]: Generating asset frames.html
[debug]: Serializing to documentation/frames.html
[debug]: Serializing to documentation/_index.html
[debug]: Serializing to documentation/index.html
[debug]: Serializing to documentation/file.README.html
[debug]: Serializing to documentation/file.README.html
[debug]: Serializing to documentation/file.README.html
[debug]: Serializing to documentation/top-level-namespace.html
Files:           0
Modules:         0 (    0 undocumented)
Classes:         0 (    0 undocumented)
Attributes:      0 (    0 undocumented)
Methods:         0 (    0 undocumented)
 100.00% documented
```
Generated documentation linking 2 of the readmes to the same content:

Screenshot 2019-09-03 at 11 04 23

## Expected Output

Expecting yard to respect the file paths and link the files to the correct content

## Environment details:

* OS: MacOs 10.14.6
* Ruby version `2.6.2`
* YARD version (`yard -v`): `0.9.20`
* Relevant software dependency/versions: none

bfad:
# Description

Currently when creating HTML documentation, extra files are stored at the web root with "file." prefixed to their name. This PR changes this practice to instead store them in their own directory (named "_file" to prevent conflicts with Ruby objects). 

This brings the output more inline with the server which hosts the files under `/file/` path. In fact, the server keeps the relative path to the extra file as part of the URL path. (For example, while yard's changelog is found at `/file/CHANGELOG.md` the getting started guide is at `/file/docs/GettingStarted.md`.) If this PR is accepted, I would like to work on doing something similar for HTML creation. (This might fix #1277 as all the READMEs would be listed there, and it's easy enough to set custom title attributes to help distinguish.) If all of that is successful, I would like to have a setting that allows for showing the extra file hierarchically similar to how nested Ruby objects are presented with disclosure triangles.

# Completed Tasks

### DIFF
--- a/lib/yard/code_objects/extra_file_object.rb
+++ b/lib/yard/code_objects/extra_file_object.rb
@@ -61,6 +61,10 @@ module YARD::CodeObjects
 
     def type; :extra_file end
 
+    def source_type; :ruby end
+
+    def parent; YARD::Registry.root end
+
     def ==(other)
       return false unless self.class === other
       other.filename == filename
--- a/lib/yard/serializers/file_system_serializer.rb
+++ b/lib/yard/serializers/file_system_serializer.rb
@@ -51,7 +51,7 @@ module YARD
         return object if object.is_a?(String)
 
         if object.is_a?(CodeObjects::ExtraFileObject)
-          fspath = ['file.' + object.name + (extension.empty? ? '' : ".#{extension}")]
+          fspath = ['_file', object.name + (extension.empty? ? '' : ".#{extension}")]
         else
           objname = object != YARD::Registry.root ? mapped_name(object) : "top-level-namespace"
           objname += '_' + object.scope.to_s[0, 1] if object.is_a?(CodeObjects::MethodObject)
--- a/lib/yard/server/doc_server_serializer.rb
+++ b/lib/yard/server/doc_server_serializer.rb
@@ -20,7 +20,7 @@ module YARD
         when CodeObjects::ConstantObject, CodeObjects::ClassVariableObject
           serialized_path(object.namespace) + "##{object.name}-#{object.type}"
         when CodeObjects::ExtraFileObject
-          super(object).gsub(/^file\./, 'file/')
+          super(object).gsub(/^_file\//, 'file/')
         else
           super(object)
         end
--- a/spec/serializers/file_system_serializer_spec.rb
+++ b/spec/serializers/file_system_serializer_spec.rb
@@ -66,7 +66,7 @@ RSpec.describe YARD::Serializers::FileSystemSerializer do
     it "handles ExtraFileObject's" do
       s = Serializers::FileSystemSerializer.new
       e = CodeObjects::ExtraFileObject.new('filename.txt', '')
-      expect(s.serialized_path(e)).to eq 'file.filename.html'
+      expect(s.serialized_path(e)).to eq '_file/filename.html'
     end
 
     it "differentiates instance and class methods from serialized path" do
--- a/spec/templates/helpers/html_helper_spec.rb
+++ b/spec/templates/helpers/html_helper_spec.rb
@@ -415,12 +415,12 @@ RSpec.describe YARD::Templates::Helpers::HtmlHelper do
       expect(parse_link(resolve_links("{file:TEST.txt#abc}"))).to eq(
         :inner_text => "TEST",
         :title => "TEST",
-        :href => "file.TEST.html#abc"
+        :href => "_file/TEST.html#abc"
       )
       expect(parse_link(resolve_links("{file:TEST.txt title}"))).to eq(
         :inner_text => "title",
         :title => "title",
-        :href => "file.TEST.html"
+        :href => "_file/TEST.html"
       )
     end
 
--- a/templates/default/fulldoc/html/setup.rb
+++ b/templates/default/fulldoc/html/setup.rb
@@ -8,7 +8,7 @@ def init
   generate_assets
   serialize('_index.html')
   options.files.each_with_index do |file, _i|
-    serialize_file(file, file.title)
+    serialize_file(file)
   end
 
   options.delete(:objects)
@@ -48,6 +48,7 @@ end
 # Generate the index document for the output
 # @params [Hash] options contains data and flags that influence the output
 def serialize_index(options)
+  options.object = Registry.root
   Templates::Engine.with_serializer('index.html', options.serializer) do
     T('layout').run(options.merge(:index => true))
   end
@@ -57,18 +58,12 @@ end
 # the README file or files specified on the command-line.
 #
 # @param [File] file object to be saved to the output
-# @param [String] title currently unused
 #
 # @see layout#diskfile
-def serialize_file(file, title = nil) # rubocop:disable Lint/UnusedMethodArgument
-  options.object = Registry.root
+def serialize_file(file)
   options.file = file
-  outfile = 'file.' + file.name + '.html'
-
   serialize_index(options) if file == options.readme
-  Templates::Engine.with_serializer(outfile, options.serializer) do
-    T('layout').run(options)
-  end
+  serialize(file)
   options.delete(:file)
 end
 
--- a/templates/guide/fulldoc/html/setup.rb
+++ b/templates/guide/fulldoc/html/setup.rb
@@ -23,7 +23,7 @@ def init
   class << options.serializer
     define_method(:serialized_path) do |object|
       if CodeObjects::ExtraFileObject === object
-        super(object).sub(/^file\./, '').downcase
+        super(object).sub(/^_file\//, '').downcase
       else
         super(object)
       end


bfad:
I'm not sure why GitHub is showing 5 different commits when it should just be the one.

lsegal:
Hi @bfad,

I'm unclear on the purpose of this change. As is, this is a pretty large breaking change, since this would change URL structure of any statically deployed documentation, which is pretty significant and would need a good justification. I get that there's an attempt to bring static generation more in line with `yard server`, but the URL structure would never be 100% compatible since `yard server` removes the suffix `.html` extensions, so without rewrites this doesn't fully achieve the result (and if URL rewrites are okay, there are rewrite rules that can achieve server parity without code changes).

If the intent is to fix #1277, there are probably other ways to do that without changing the naming scheme on disk. Presumably this could be done with similar custom file metadata attributes to `@title`, like `@filename` to rename a file when serializing.

bfad:
@lsegal thanks for looking this over. 

The ultimate end goal is the ability to have a hierarchical structure for these extra documents. The idea being that I would use globing and the file hierarchy would be replicated, not just at the file-system level, but also with disclosure triangles in the HTML sidebar (just like namespace hierarchies).

This could still be done with something like a `@docpath` attribute, but if I have already done the work to organize the documents into folders, then it seems onerous to have me spell out that hierarchy for each file in a metadata attribute. At the same time, I really respect your hesitation to break existing URLs. If you don't think this is the right path to go down, I can investigate alternatives.

lsegal:
Backing up a bit, it seems like we're solving for two separate problems.

The location that YARD uses to serialize files on disk, as well as the expected URL structure by `yard server`, are actually quite a bit disconnected from the way extra files are rendered in templates, in other words, whether they are displayed flat or in hierarchy, what title is used to display the links, etc.

YARD actually stores the full location to each extra file alongside the file object inside the `filename` attribute; `.name` is sanitized and stripped but the filename is untouched. Even though those aren't used in serialization, they _could_ certainly be used in the default (or custom) HTML generation template, if there was value to copying the hierarchy when displaying extra files. I think there is still a question of whether this is useful, or what heuristic YARD should use when doing so (for ex: YARD's use of `doc/X` shouldn't display as a `> doc/` expandable item, there's probably a basic nesting heuristic that could be applied to avoid this), but this is certainly solvable with the information already available in YARD's cache database and [only requires template changes](https://github.com/lsegal/yard/blob/27874ee1532b00351722cd9e3ce4cf0822768916/templates/default/fulldoc/html/full_list_file.erb#L4) (also [here](https://github.com/lsegal/yard/blob/27874ee1532b00351722cd9e3ce4cf0822768916/templates/default/layout/html/files.erb#L6)).

Now, duplicate/conflicting filenames from subdirs is a separate problem that still exists even if you solve the above. That said, the location that YARD chooses to serialize files to should _generally_ not be a user concern and should typically be hidden from the user, so realistically, there are a few options to disambiguate conflicting filenames that could work without a breaking change or even introducing an attribute:

1. Perform a duplicate check against extra files prior to serializing and rewrite the `.name` attribute to use `.filename` if dupes are found. This is backward compatible since the old behavior will be to continue using flattened file paths. This also doesn't change anything in the serializer itself and can be done fully [inside of the template](https://github.com/lsegal/yard/blob/27874ee1532b00351722cd9e3ce4cf0822768916/templates/guide/fulldoc/html/setup.rb#L47-L65). Alternatively you could do this dupe checking inside of [YARD::CLI::Yardoc](https://github.com/lsegal/yard/blob/27874ee1532b00351722cd9e3ce4cf0822768916/lib/yard/cli/yardoc.rb#L412-L419). The serializer should be able to handle this case out of the box.
2. Optionally you could add an extra attribute to ExtraFileObject (like `.alt_path`) that is used by `.path` if present in order to override the serialization location. This is less nice because of all the hoops you'd need to jump through and extra complexity, but it would retain all original state information (`.name` won't be rewritten out from under you).

Hope this helps / makes sense.

lsegal:
Sidenote: these can and should be two separable and independent PRs. The latter, specifically could be accepted to fix the underlying issue of dupes without needing to change how files render in the template.

lsegal:
Here's a first-attempt patch for supporting dupe files:

```diff
diff --git a/lib/yard/cli/yardoc.rb b/lib/yard/cli/yardoc.rb
index 5332196f..8a6804e1 100644
--- a/lib/yard/cli/yardoc.rb
+++ b/lib/yard/cli/yardoc.rb
@@ -222,6 +222,7 @@ module YARD
         @save_yardoc = true
         @has_markup = false
         @fail_on_warning = false
+        @extra_file_dupes = {}

         if defined?(::Encoding) && ::Encoding.respond_to?(:default_external=)
           utf8 = ::Encoding.find('utf-8')
@@ -413,7 +414,17 @@ module YARD
         files.map! {|f| f.include?("*") ? Dir.glob(f) : f }.flatten!
         files.each do |file|
           if extra_file_valid?(file)
-            options.files << CodeObjects::ExtraFileObject.new(file)
+            extra_file = CodeObjects::ExtraFileObject.new(file)
+            options.files << extra_file
+
+            if @extra_file_dupes[extra_file.name]
+              extra_file.attributes[:title] ||= extra_file.filename
+              extra_file.name = extra_file.filename.
+                gsub(/\.[^.]+$/, '').
+                gsub(/\//, '_')
+            else
+              @extra_file_dupes[extra_file.name] = true
+            end
           end
         end
       end
```

The gsub stuff could be refactored into ExtraFileObject's name= setter to remove the extension, as well as a change to FileSystemSerializer to correctly escape `/` characters for ExtraFileObjects-- the latter would remove the need for the `.attributes[:title]` override.

bfad:
What if, instead of messing with `ExtraFileObject#name`, we used `ExtraFileObject#path` for this? Patch:
```diff
diff --git a/lib/yard/code_objects/extra_file_object.rb b/lib/yard/code_objects/extra_file_object.rb
index c0f53a3e..eb9bad6f 100644
--- a/lib/yard/code_objects/extra_file_object.rb
+++ b/lib/yard/code_objects/extra_file_object.rb
@@ -8,6 +8,7 @@ module YARD::CodeObjects
     attr_accessor :filename
     attr_writer :attributes
     attr_accessor :name
+    attr_accessor :path
     # @since 0.8.3
     attr_reader :locale
 
@@ -18,6 +19,7 @@ module YARD::CodeObjects
     def initialize(filename, contents = nil)
       self.filename = filename
       self.name = File.basename(filename).gsub(/\.[^.]+$/, '')
+      self.path = filename.gsub(/\.[^.]+$/, '')
       self.attributes = SymbolHash.new(false)
       @original_contents = contents
       @parsed = false
@@ -25,8 +27,6 @@ module YARD::CodeObjects
       ensure_parsed
     end
 
-    alias path name
-
     def attributes
       ensure_parsed
       @attributes
diff --git a/templates/default/fulldoc/html/setup.rb b/templates/default/fulldoc/html/setup.rb
index 201ba63c..bd0e1954 100644
--- a/templates/default/fulldoc/html/setup.rb
+++ b/templates/default/fulldoc/html/setup.rb
@@ -7,8 +7,11 @@ def init
   return serialize_onefile if options.onefile
   generate_assets
   serialize('_index.html')
+
+  dup_file_names = {}
   options.files.each_with_index do |file, _i|
-    serialize_file(file, file.title)
+    serialize_file(file, dup_file_names[file.name] ? file.path : file.name)
+    dup_file_names[file.name] = true
   end
 
   options.delete(:objects)
@@ -60,10 +63,10 @@ end
 # @param [String] title currently unused
 #
 # @see layout#diskfile
-def serialize_file(file, title = nil) # rubocop:disable Lint/UnusedMethodArgument
+def serialize_file(file, file_name)
   options.object = Registry.root
   options.file = file
-  outfile = 'file.' + file.name + '.html'
+  outfile = 'file.' + file_name.gsub(/\//, '_') + '.html'
 
   serialize_index(options) if file == options.readme
   Templates::Engine.with_serializer(outfile, options.serializer) do
```

IRod22:
@lsegal @bfad Hello. If I may ask, when will this PR get merged? I have some [documentation](https://irod22.github.io/fedi-joke-bot/file.README.html) for a pet project I'm working on, and the docs have some markdown files that link to each other, but those links are broken; one of the reasons they are broken is because YARD's flat file structure is incompatible with the file/folder structure the browser expects. I'm hoping this gets merged ASAP because I released V1 of my project with broken documentation [because I knew that a fix would not come soon](https://urusai.social/@ivan18rod/110363179636169949).

lsegal:
> [!NOTE]

PR #1290, opened 2019-11-19
ruby_parser.rb - add void_stmt to insert_comments loop

MSP-Greg:
# Description

While running YARD with Ruby 2.7, the following comment is not added to class B:
```ruby
module A
  # comment
  # ...
  # end comment
  class B
```

but, if a blank line is added between `module A` and the comment, it parses fine.

Tests passed in my fork, but 2.7/trunk is not tested against.  I checked the issue locally.

I could add Actions testing on Windows (Ruby 2.4 thru trunk), which I assume would show the issue.

# Completed Tasks

* [x] I have read the [Contributing Guide][contrib].
* [x] The pull request is complete (implemented / written).
* [x] Git commits have been cleaned up (squash WIP / revert commits).
* [ ] I wrote tests and ran `bundle exec rake` locally (if code is attached to PR).

[contrib]: https://github.com/lsegal/yard/blob/master/CONTRIBUTING.md

PR #1296, opened 2019-12-26
Support Ruby 2.7.0

koic:
# Description

Follow up of https://github.com/lsegal/yard/pull/1290#issuecomment-558204534.

This is a workaround path for Ruby 2.7.0.

`irb/slex` internal API has been removed in the following commits.
https://github.com/ruby/ruby/commit/45bb6f28db04a1f267f5b8d79392cb35087510d8

This PR copies the above file to lib/yard/parser/ruby/legacy/irb/slex.rb.

Hopefully this workaround patch will be dropped along with legacy Ruby code in the future.

And E2MM (e2mmap gem) is no longer bundled gems.
https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/

```console
% cd path/to/repo/yard
% ruby -v
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-darwin17]
% bundle exec rake

(snip)

LoadError:
  cannot load such file -- e2mmap
  # ./lib/yard/parser/ruby/legacy/ruby_lex.rb:1:in `require'
  # ./lib/yard/parser/ruby/legacy/ruby_lex.rb:1:in `<top (required)>'
```

This PR adds `e2mmap` gem to the gemspec. And this dependency on `e2mmap` could be removed when `irb/slex` is removed in the future.

# Completed Tasks


yard.gemspec
```diff
@@ -21,4 +21,7 @@ Gem::Specification.new do |s|
   s.executables   = ['yard', 'yardoc', 'yri']
   s.license = 'MIT' if s.respond_to?(:license=)
   s.metadata['yard.run'] = 'yri'
+
+  # FIXME: The following dependency can be removed when dropping legacy Ruby support.
+  s.add_runtime_dependency('e2mmap')
```
lsegal: e2mmap isn't needed, the library isn't actually used anywhere in the file. We can remove the require call instead. Tests still pass without it for me.

lsegal: We can probably get rid of this dep with a small change in slex since it's only used in a few lines of easily replaceable code.

lsegal:
Thanks for this PR!

koic:
Thanks too!

pirj:
Thanks a lot guys!

PR #1298, opened 2019-12-31
Revert "Bump vendored jquery version to 3.4.1."

lsegal:
Still needs some work to update usages


skalee:
If the problem is related to [browser support](https://jquery.com/browser-support/), then perhaps jQuery should be bumped to 1.12.4 instead, which is the latest of 1.x branch. Still old ([2016](https://blog.jquery.com/2016/05/20/jquery-1-12-4-and-2-2-4-released/)), but much newer than 1.7.1 ([2011](https://blog.jquery.com/2011/11/21/jquery-1-7-1-released/)).

lsegal:
I think this has more to do with breaking API changes, though 1.12.4 might help with that too.

thomthom:
GitHub is yielding security reports due to jQuery 1.7 for our repo hosting our Ruby API documentation:

image

Upgrading to the latest 1.x version in YARD would be good - avoiding the need to patch the output downstream.

thomthom:
hmm... With 1.12 and 1.9 I see rendering issues:

2020-10-21_12h21_26

1.8 looks fine. But that's not new enough for the security alerts.

thomthom:
Ah, this related to 1.9 removing a number of functions, such as `toggle(function, function)`: https://jquery.com/upgrade-guide/1.9/#toggle-function-function-removed

By adding the migration plugin it worked fine again. Though at the cost of another JS to request. Unless it's bundled into a single file...

PR #1301, opened 2020-01-10
Fixing yard server parsing checksum containing filenames with spaces.

vadviktor:
# Description

Running `yard server` it parses the checksums file. If any files scanned by yard contains space in their name, the parsing of the checksum file fails.

Similarly reported in:
https://stackoverflow.com/questions/27320253/yard-server-yields-invalid-number-of-elements-3-for-1-2

# Completed Tasks

### DIFF
--- a/lib/yard/registry_store.rb
+++ b/lib/yard/registry_store.rb
@@ -291,7 +291,8 @@ module YARD
     def load_checksums
       return unless File.file?(checksums_path)
       lines = File.readlines(checksums_path).map do |line|
-        line.strip.split(/\s+/)
+        parts = line.strip.split(/\s+/)
+        parts.count > 2 ? [parts[0..-2].join(' '), parts[-1]] : parts
       end
       @checksums = Hash[lines]
     end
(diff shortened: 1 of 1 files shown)


vadviktor:
@lsegal I'll try to figure out a test a bit later.

lib/yard/registry_store.rb
```diff
@@ -291,7 +291,8 @@ def load_object_types
     def load_checksums
       return unless File.file?(checksums_path)
       lines = File.readlines(checksums_path).map do |line|
-        line.strip.split(/\s+/)
+        parts = line.strip.split(/\s+/)
+        parts.count > 2 ? [parts[0..-2].join(' '), parts[-1]] : parts
```
lsegal: Might be simpler to use:

```suggestion
line.strip.rpartition(' ').tap {|p| p.delete_at(1) }
```

(replaces the first line too, GitHub doesn't support multiline suggestions yet)

vadviktor: I like this approach better too, thanks for pointing it out. 👍

lsegal:
Thanks for the PR!

PR #1307, opened 2020-02-03
Omit spec files when building gem

dduugg:
# Description

This reduces the gem size by ~20% (902k->738k) by omitting spec files (there may be further space optimizations in this style, this PR doesn't attempt to be comprehensive). This is inspired by the default implementation when creating a gem via `bundle gem`, which is:
```
`git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
```

I've also cleaned up `.rubocop.yml` to resolve local plugin errors. (Let me know if I've overstepped the PR scope and i'll cut it.)

# Completed Tasks


lsegal:
Thanks for the fix!

PR #1318, opened 2020-03-26
Update rdoc_markup.rb

ioquatix:
Don't lazy initialize mutex, ensure mutex is locked when updating associated state.

### DIFF
--- a/lib/yard/templates/helpers/markup/rdoc_markup.rb
+++ b/lib/yard/templates/helpers/markup/rdoc_markup.rb
@@ -34,16 +34,17 @@ module YARD
         class RDocMarkup
           attr_accessor :from_path
 
+          @@mutex = Mutex.new
           @@formatter = nil
           @@markup = nil
-          @@mutex = nil
 
           def initialize(text)
             @text = text
 
-            @@formatter ||= RDocMarkupToHtml.new
-            @@markup ||= MARKUP.new
-            @@mutex ||= Mutex.new
+            @@mutex.synchronize do
+              @@formatter ||= RDocMarkupToHtml.new
+              @@markup ||= MARKUP.new
+            end
           end
 
           def to_html


lib/yard/templates/helpers/markup/rdoc_markup.rb
```diff
@@ -34,16 +34,17 @@ class RDocMarkupToHtml < SM::ToHtml; end
...
 
           def initialize(text)
             @text = text
 
-            @@formatter ||= RDocMarkupToHtml.new
-            @@markup ||= MARKUP.new
-            @@mutex ||= Mutex.new
+            @@mutex.synchronize do
+              @@formatter ||= RDocMarkupToHtml.new
+              @@markup ||= MARKUP.new
```
headius: Why are these lazy to begin with?

ioquatix:
Good question.

lib/yard/templates/helpers/markup/rdoc_markup.rb
ioquatix: Good question!

lsegal: @headius to minimize memory overhead when the classes might be autoloaded but not necessarily used. Since RDocMarkupToHtml is not owned by YARD, there's no reliable way to control how much memory is being wired up when it's initialized. 

That said, the case where this file is autoloaded but not used is on the unlikely side, so if there are good reasons to change it, I'm not opposed. I don't see a problem with the change as-is, though.

ioquatix: Is it so slow that making one instance per class is out of the question?

lsegal: Sorry for the late reply @ioquatix: I'm not sure, but these things are instantiated quite often so I would guess the answer is yes. If you want to benchmark to check it out, feel free!

lsegal:
Going to take this as is and if we need to touch anything up we can do it in a subsequent PR.

ioquatix:
Great.

PR #1320, opened 2020-04-06
Fix rdoc generation for Ruby 1.9 variants

mullermp:
# Description

Yard supports Ruby > 0 but rdoc documentation generation fails for Ruby 1.9. `%i` symbols didn't exist in Ruby 1.9 so this change uses an array of symbols explicitly instead. Also, the `Range` class in Ruby 1.9 did not have a `#size` method, so where we are checking size on those, convert it to an array first. Unfortunately some hard to migrate Amazon internal tooling still uses old Ruby variants. For testing, I ran `bundle exec rake` for 2.6.5, and internally, I tested that documentation is generated for Ruby 1.9 packages.

# Completed Tasks

### DIFF
--- a/lib/yard/parser/ruby/ruby_parser.rb
+++ b/lib/yard/parser/ruby/ruby_parser.rb
@@ -412,7 +412,7 @@ module YARD
         end
 
         def on_dyna_symbol(sym)
-          rng = if sym.source_range.size == 0 # rubocop:disable Style/ZeroLengthPredicate
+          rng = if sym.source_range.to_a.size == 0 # rubocop:disable Style/ZeroLengthPredicate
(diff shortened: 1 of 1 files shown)


mullermp:
If the gemspec continues to support Ruby > 0 I suggest having some testing for older versions too :)

mullermp:
friendly ping on this @lsegal 

Also, was surprised to see your alias pop up in some internal wikis! Wow! Did not know that you worked here a long time ago. BTW -- Blackfoot is still a building, and it sucks. X)

lsegal:
Thanks for the PR! Glad to see our building still stands :)

mullermp:
Trevor (my manager) says hello. :)

PR #1323, opened 2020-04-11
Add `:with_toc_data` to redcarpet configs

olbrich:
This will autogenerate id tags for headers. Without headers the javascript in the default template will generate ids, but uses a different algorithm that doesn't match what github does. The javascript will leave existing ids alone, so if the default redcarpet markdown renderer is used the ids will match what github does and it will be possible to reference anchor tags in a way that works both on github and when rendered via a yard server.

# Description

# Completed Tasks


esotericpig:
While it fixes it for most GitHub Markdown links, it still breaks for all non-English and special circumstances:

```
Name / WithSlash
日本語
中文
```

Non-English links become `part-edb592...`

So it still doesn't use GitHub's algorithm, but uses RedCarpet's pseudo-GitHub algorithm.

PR #1331, opened 2020-05-12
Add .gitattributes with line-ending rule

olleolleolle:
# Description

  - avoids "EOL issue, ie CRLF vs LF" by adding a .gitattributes file with a rule
  - see #1328

[GitHub article on line endings & git](https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings).

# Completed Tasks

* [x] I have read the [Contributing Guide][contrib].
* [x] The pull request is complete (implemented / written).
* [x] Git commits have been cleaned up (squash WIP / revert commits).
* [ ] ~I wrote tests and ran `bundle exec rake` locally (if code is attached to PR).~

[contrib]: https://github.com/lsegal/yard/blob/master/CONTRIBUTING.md

PR #1345, opened 2020-07-18
Fix line endings

adam12:
Force the filetype of `png` files to be binary, preventing git from assuming they've changed.

Converted `lib/yard/handlers/c/base.rb` line endings to LF. It's somewhat noisy but allows for a clean checkout of the git tree. Hopefully the fixes in #1331 will prevent CRLF from landing in the repo again.

This fixes a dirty tree on clone (Linux).

# Completed Tasks


lsegal:
Thanks for the quick fix!

PR #1346, opened 2020-07-28
Prevent footer from being partially hidden on Firefox

benoittgt:
# Description

I noticed that the footer is partially hidden in Firefox but not in Chrome. This patch make the footer fully visible.

Jul-28-2020 14-41-14

# Completed Tasks


lsegal:
Thanks for the quick fix!

Issue #1376, opened 2021-04-09
`Unhandled exception in YARD::Handlers::Ruby::MethodHandler` when using Ruby 3.0' endless method definitions


koic:
YARD 0.9.26 occurs an error when trying to parse Ruby 3.0 endless method definitions.

## Steps to reproduce

```console
% cat endless_method_definition.rb
def say = puts('hello')
```

```console
% yard doc endless_method_definition.rb
[error]: Unhandled exception in YARD::Handlers::Ruby::MethodHandler:
  in `endless_method_definition.rb`:1:

        1: def say = puts('hello')

Files:           0
Modules:         0 (    0 undocumented)
Classes:         0 (    0 undocumented)
Constants:       0 (    0 undocumented)
Attributes:      0 (    0 undocumented)
Methods:         0 (    0 undocumented)
 100.00% documented
```

## Actual Output

```console
% yard doc --debug endless_method_definition.rb
[debug]: Parsing ["endless_method_definition.rb"] with `ruby` parser
[debug]: Parsing endless_method_definition.rb
[error]: Unhandled exception in YARD::Handlers::Ruby::MethodHandler:
  in `endless_method_definition.rb`:1:

        1: def say = puts('hello')

[error]: NoMethodError: undefined method `type' for nil:NilClass
[error]: Stack trace:
        /Users/koic/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/yard-0.9.26/lib/yard/parser/ruby/ast_node.rb:484:in `parameters'
        /Users/koic/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/yard-0.9.26/lib/yard/handlers/ruby/method_handler.rb:70:in `format_args'
        /Users/koic/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/yard-0.9.26/lib/yard/handlers/ruby/method_handler.rb:10:in `block in <class:MethodHandler>'
        /Users/koic/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/yard-0.9.26/lib/yard/handlers/processor.rb:114:in `block (2 levels) in process'
        /Users/koic/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/yard-0.9.26/lib/yard/handlers/processor.rb:112:in `each'
[...]

PR #1381, opened 2021-05-06
Support endless method definitions

rkh:
# Description

Ruby 3.0 introduces endless method definitions:

``` ruby
def method = "return value"
```

This raises an exception in YARD, as the Ruby parser never calls `on_params`, thus there is no `ParameterNode` inside the method definition (see #1376). I've adjusted the code slightly to be able to handle this case.

I've added a test and have also confirmed on a real world project that with this change YARD correctly handles endless method definitions without parameters.

Endless method definitions with parameters were already supported. I've also added a test for them, as there was none.

# Completed Tasks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # main
 
+- Add support for Ruby 3.0 endless method definitions. (#1376)
+
 # 0.9.26 - December 26th, 2020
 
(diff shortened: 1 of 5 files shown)


rkh:
The coverage report is a false positive, as the lines in question are only executed with Ruby 3.0 or higher.

lsegal:
Thanks for your PR! Sometimes coverage acts up on small deltas.

PR #1389, opened 2021-06-17
Fix nested array parsing

dduugg:
# Description

Prior to this change, nested arrays would be missing closing brackets when translating back to source code, e.g.:
```
[1] pry(main)> YARD::Parser::Ruby::RubyParser.new('a[b[c]]', nil).parse.root.first.source
=> "a[b[c]"
```

This in turn would print warnings in YARD output, and result in documentation errors. Example warnings:
```
[warn]: Syntax error in `(stdin)`:(1,96): syntax error, unexpected end-of-input, expecting ']'
[warn]: ParserSyntaxError: syntax error in `(stdin)`:(1,96): syntax error, unexpected end-of-input, expecting ']'
```

This change corrects this specific problem. All tests pass, but this is a naïve fix, and I'm not familiar enough with this code to contemplate other potential side effects.

# Completed Tasks

### DIFF
--- a/lib/yard/parser/ruby/ruby_parser.rb
+++ b/lib/yard/parser/ruby/ruby_parser.rb
@@ -370,7 +370,7 @@ module YARD
 
         def on_aref(*args)
           @map[:lbracket].pop
-          ll, lc = *@map[:aref].pop
+          ll, lc = *@map[:aref].shift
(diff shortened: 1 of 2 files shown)


lsegal:
Sorry it took so long to get to this. Thanks for the fix!

PR #1391, opened 2021-07-19
better import

noraj:
# Description

# Completed Tasks

### DIFF
--- a/yard.gemspec
+++ b/yard.gemspec
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-require File.expand_path('../lib/yard/version', __FILE__)
+require_relative 'lib/yard/version'
 
 Gem::Specification.new do |s|
   s.name          = "yard"


yard.gemspec
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-require File.expand_path('../lib/yard/version', __FILE__)
+require_relative 'lib/yard/version'
```
lsegal, line 2: Is there a specific need to change this? Compatibility? Bug fix?

noraj: Chaging a very cumbersome and difficult to understand syntax into something much more easier to understand.
require_relative is available since at least ruby 2.1 so it's no issue for compatibility in most projects, I didn't found the ruby version supported by yard.

lsegal: YARD is compatible back to Ruby 1.8. Although we don't actively support older versions, we still need a specific reason to make a change. 

Understandability could be a reason, but in this particular case, the new syntax is not much more understandable than the original, especially since the changed require syntax is used in many other places in the YARD codebase, as well as many other Ruby projects. The `expand_path(path, __FILE__)` idiom is still an actively used one; not always for require, but nonetheless, even [Rails has this syntax in its gemspec](https://github.com/rails/rails/blob/main/rails.gemspec#L3).

noraj: Ok if yard want retro-compatibility for older ruby version this should be avoided

PR #1419, opened 2022-01-01
Only call Object#taint when the taint method exists

jcouball:
# Description

Running YARD with Ruby 3.2 (current Ruby HEAD) results in the error:

```
$ rake yard
NoMethodError: undefined method `taint' for "# frozen_string_literal: true\n...":String

            module_eval(File.read(setup_file).taint, setup_file, 1)
                                             ^^^^^^

/home/runner/work/ruby-git/ruby-git/vendor/bundle/ruby/3.2.0/gems/yard-0.9.27/lib/yard/templates/template.rb:179:in `load_setup_rb'
```

This is because the `taint` mechanism, which was deprecated in Ruby 2.7 is to be removed from Ruby 3.2.  In Ruby 2.7, the taint checking functionality was removed and `Object#taint`, `Object#trust`, `Object#untaint`, `Object#untrust` and related methods have no-op implementations. This means all objects by default are deemed untainted.

In Ruby 3.2 (the current HEAD), `Object#taint`, `Object#trust`, `Object#untaint`, `Object#untrust` are completely removed.

See the article [Ruby 2.7 removes taint checking mechanism](https://blog.saeloun.com/2020/02/18/ruby-2-7-access-and-setting-of-safe-warned-will-become-global-variable.html) for a summary of the timeline for removal of the taint mechanism and links to Ruby core team discussions.

# Caveats

I ran tests locally after making these changes in Ruby 2.4.x so the CI matrix build will need to be completed to make sure that these changes do not impact tests on all supported versions of Ruby.

# Completed Tasks

- [X] I ran `bundle exec rake` locally (if code is attached to PR).

### DIFF
--- a/lib/yard/cli/stats.rb
+++ b/lib/yard/cli/stats.rb
@@ -225,7 +225,8 @@ module YARD
         end
 
         opts.on('--query QUERY', "Only includes objects that match a specific query") do |query|
-          options[:verifier].add_expressions(query.taint)
+          query.taint if query.respond_to?(:taint)
+          options[:verifier].add_expressions(query)
         end
       end
     end
--- a/lib/yard/cli/yardoc.rb
+++ b/lib/yard/cli/yardoc.rb
@@ -650,7 +650,8 @@ module YARD
 
         opts.on('--query QUERY', "Only show objects that match a specific query") do |query|
           next if YARD::Config.options[:safe_mode]
-          options.verifier.add_expressions(query.taint)
+          query.taint if query.respond_to?(:taint)
+          options.verifier.add_expressions(query)
         end
 
         opts.on('--title TITLE', 'Add a specific title to HTML documents') do |title|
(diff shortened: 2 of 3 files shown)


lsegal (commented):
We would still want to support taint for Ruby versions that support it.

jcouball:
@lsegal I changed the implementation to call `taint` when the receiving object `respond_to?(:taint)`. Can you take another look?

lsegal (approved):
Great stuff! Thanks so much for this PR!

ShockwaveNN:
@lsegal Any chance you will release this change soon?

Really want to test my CI with `ruby-head`, but this fix is required (

jcouball:
@ShockwaveNN here is what I did as a temporary workaround until a new YARD version is published: ruby-git/ruby-git#573

ShockwaveNN:
@jcouball Thanks, not a big fan of installing something from master )

I'm interesting if there is some specific problem why owner cannot release (something like v0.9.28) with this fix

jcouball:
@ShockwaveNN 100% agree, but I was tired of failing builds. 

While I understand that YARD is only a dev dependency and Ruby 3.2 isn't released, fingers crossed that @lsegal will make a release soon with this fix.

PR #1430, opened 2022-01-31
Add Ruby 3.1 and ruby-head to CI

petergoldstein:
# Description

This PR adds Ruby 3.1 and ruby-head to CI.

As part of that change it fixes a few keyword argument issues in the specs - three specs were failing because the arguments passed to `with` were being intepreted as keyword arguments.  Wrapping them in explicit hashes resolved those failures.

I also observed a pre-existing failure to post to Coveralls (for all Rubies) that I did not address.

# Completed Tasks

- [X] I wrote tests and ran `bundle exec rake` locally (if code is attached to PR).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       matrix:
         os: [ubuntu-20.04, macos-10.15, windows-2019]
         # 3.0 is interpreted as 3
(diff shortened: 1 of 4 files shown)


lsegal:
This was resolved by #1424 and #1433, but thank you for the contribution!

Note that ruby-head will not be added since we historically had gottten a lot of regression based failures that I do not have enough bandwidth to work around.

PR #1431, opened 2022-02-01
Support ruby 3.1 :args_forward shape change

nvasilevski:
# Description

Currently yard errors when tries to parse method with args forwarding using Ruby 3.1:
```
[error]: NoMethodError: undefined method `source' for "&":String

params << ['&' + args.block_param.source, nil] if args.block_param
  | [error]: Stack trace:
  | /tmp/bundle/ruby/3.1.0/gems/yard-0.9.27/lib/yard/handlers/ruby/method_handler.rb:100:in `format_args'
```

This is happening because Ruby 3.1 changes `params` statement for a method with args forwarding from:
Reference:
https://kddnewton.com/2022/01/08/ripper-changelog-3.1.0.html#argument-forwarding

```ruby
[:params, nil, nil, [:args_forward], nil, nil, nil, nil]
```
to:
```ruby
 [:params, nil, nil, nil, nil, nil, [:args_forward], :&]
```

So the `block_param` wrongly assumes that `:&` is an explicit block argument.
I don't think there is an issue with `block_param` method, the "shape" of the statement stays the same and the last value in the array is references block param, it's just in this particular case the block param is "anonymous" or "implicit" if I may call it like that.

So I'm adding additional `args_forward`  check to avoid processing `:&` block.
`args_forward` method works for both Ruby 3.1 and older versions even though we don't necessarily need it for Ruby lower than 3.1

Also added test for the anonymous block forwarding. It's not related to the issue but I thought it can't hurt having it just in case

# Completed Tasks

### DIFF
--- a/lib/yard.rb
+++ b/lib/yard.rb
@@ -51,6 +51,9 @@ module YARD
 
   # @return [Boolean] whether YARD is being run in Ruby 3.0
   def self.ruby3?; @ruby3 ||= (RUBY_VERSION >= '3.0.0') end
+
+  # @return [Boolean] whether YARD is being run in Ruby 3.1
+  def self.ruby31?; @ruby31 ||= (RUBY_VERSION >= '3.1.0') end
 end
 
 # Keep track of Ruby version for compatibility code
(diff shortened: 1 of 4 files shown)


lsegal:
Thanks for the implementation! Sorry for the delay on this, it looks good!

PR #1449, opened 2022-08-09
gemspec: Avoid including .rubocop.yml

olleolleolle:
# Description

This PR removes a (1) file from the gemspec's `files` directive.

With default settings, a consuming project could inadvertently end up including the directives set up for the Yard project.

Example: a setup where there's a temporary directory with gems inside the app.

```
Error: The `Style/MethodMissingSuper` cop has been removed since it has been superseded by `Lint/MissingSuper`. Please use `Lint/MissingSuper` instead.
(obsolete configuration found in app/views/jasmine/tmp/devbox/gems_inside_the_current_ruby_install/gems/yard-0.9.27/.rubocop.yml, please update it)
```

# Completed Tasks

### DIFF
--- a/yard.gemspec
+++ b/yard.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.email         = "lsegal@soen.ca"
   s.homepage      = "http://yardoc.org"
   s.platform      = Gem::Platform::RUBY
-  s.files         = `git ls-files`.strip.split(/\s+/).reject {|f| f.match(%r{^spec/}) }
+  s.files         = `git ls-files`.strip.split(/\s+/).reject {|f| f.match(%r{^spec/}) || f == '.rubocop.yml' }
   s.require_paths = ['lib']
   s.executables   = ['yard', 'yardoc', 'yri']
   s.license = 'MIT' if s.respond_to?(:license=)


lsegal:
I mostly think that removing the file should be okay, but I'm not sure about the specific motivation for this change.

In the case you've described, Rubocop is (helpfully) telling you that it's running against a source tree that it probably shouldn't be. Put another way, if Rubocop is going to be running against YARD source (which it is in this example), it should probably at least be following consistent linter rules for the source tree so as not to accidentally autofix and reformat a bunch of code.

If this is just about the specific `Style/MethodMissingSuper` rule from logging a messy warning to the user, then it seems like we should probably be addressing that directly, which this PR does not do. In other words, it kind of seems to me like the intent of this PR is to silence a warning rather than fix the root cause. The better fix would be to update the lint rule IMO.

pirj:
It would make sense to reduce the gem size by removing files not needed for the runtime, e.g.:
 - benchmarks
 - docs
 - po/ja.po
 - .github/*
 - Rakefile
 - Gemfile
 - .rspec
 - ...

lsegal:
This was pulled from #1480, but thank you for starting the conversation!

PR #1451, opened 2022-08-18
Avoid extra allocations when checking for namespace separators

ghost:
# Description

Hi, I am trying to optimize YARD a bit to speed up documentation runs on a large codebase.

The codebase is private, so I am using https://github.com/watir/watir/ as a smaller stand-in to report benchmarking numbers from https://github.com/SamSaffron/memory_profiler.

# Test set-up
- Ruby 2.7.6 (installed via `rvm` on Linux)
- Watir version: 293bed2b57c
- MemoryProfiler version: 1.0.0
- YARD baseline: 0a550939f9b
- Test command:
  - `ruby-memory-profiler --no-color --retained-strings=500 --allocated-strings=500 --max=300 -o prof.log ./run-yard.rb stats`
    - `run-yard.rb` is a smaller version of `/bin/yard` (MemoryProfiler was not able to run `/bin/yard`)
    - I am using the `stats` command to avoid memory exhaustion

# Performance data
- Before
  - Total allocated: 187901758 bytes (1987572 objects)
  - Total retained:  4052113 bytes (53463 objects)
- After
  - Total allocated: 180313268 bytes (1942822 objects)
  - Total retained:  4052153 bytes (53464 objects)
- Savings
  - Total allocated: 7588490 bytes (44750 objects)
  - Total retained:  -40 bytes (-1 object)

# Completed Tasks

### DIFF
--- a/lib/yard/code_objects/proxy.rb
+++ b/lib/yard/code_objects/proxy.rb
@@ -34,7 +34,7 @@ module YARD
       def initialize(namespace, name, type = nil)
         namespace = Registry.root if !namespace || namespace == :root
 
-        if name =~ /^#{NSEPQ}/
+        if name[0, 2] == NSEP
           namespace = Registry.root
           name = name[2..-1]
         end
@@ -75,8 +75,8 @@ module YARD
         # If the name begins with "::" (like "::String")
         # this is definitely a root level object, so
         # remove the namespace and attach it to the root
-        if @name =~ /^#{NSEPQ}/
-          @name.gsub!(/^#{NSEPQ}/, '')
+        if @name[0, 2] == NSEP
+          @name = @name[2..-1]
           @namespace = Registry.root
         end
       end


lib/yard/code_objects/proxy.rb
```diff
@@ -34,7 +34,7 @@ def self.===(other) other.is_a?(self) end
       def initialize(namespace, name, type = nil)
         namespace = Registry.root if !namespace || namespace == :root
 
-        if name =~ /^#{NSEPQ}/
+        if name[0, 2] == NSEP
```
lsegal, line 37: A generalized implementation would require you to check up to `NSEP.size`:

```suggestion
        if name[0, NSEP.size] == NSEP
```

lib/yard/code_objects/proxy.rb
```diff
@@ -75,8 +75,8 @@ def initialize(namespace, name, type = nil)
         # If the name begins with "::" (like "::String")
         # this is definitely a root level object, so
         # remove the namespace and attach it to the root
-        if @name =~ /^#{NSEPQ}/
-          @name.gsub!(/^#{NSEPQ}/, '')
+        if @name[0, 2] == NSEP
+          @name = @name[2..-1]
```
lsegal, line 79: Same as above:

```suggestion
        if @name[0, NSEP.size] == NSEP
          @name = @name[NSEP.size..-1]
```

lsegal (changes requested):
This feels like a brittle change. NSEP is not guaranteed to be 2 characters, and thus, any changes away from a Regex would require a more generalized string comparison.

Even though this might seem like the `::` separation is fairly baked into the Ruby language, YARD does allow for extensions to modify the source types being parsed ([yard-js](https://github.com/lsegal/yard-js) is one example), and this would break those plugins.

It's a pretty minor change (suggestions already provided) but I'm wondering if this significantly affects the performance improvements?

PR #1457, opened 2022-09-01
Make the rake task depend only on rake

elia:
# Description

Without this it's necessary to require the whole library just to have
the task defined. With this change `yard` will be required before
executing the task code reducing the load time and memory footprint
of libraries defining the task.

# Completed Tasks

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ The second most obvious is to generate docs via a Rake task. You can do this by
 adding the following to your `Rakefile`:
 
 ```ruby
-require 'yard'
+require 'yard/rake/yardoc_task'
 
 YARD::Rake::YardocTask.new do |t|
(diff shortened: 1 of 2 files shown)


lsegal (commented):
I'm generally worried that this advice _may_ break plugins that rely on YARD's plugin loading features.

YARD relies heavily on autoload, and thus generally loads fairly lazily (minus the above plugin caveat), so generally speaking, the intended and recommended way to require YARD is to do so via `require 'yard'`, pulling in any specific classes via autoload. The location of specific Ruby sources within the YARD package are generally not considered part of our "public API", and thus, requiring the YardocTask directly in this way could break if we ever moved the file.

Is there a reason why you think this is needed? Is there a specific performance issue that necessitates this change?

lsegal:
Marking this as closed and not needed.

PR #1459, opened 2022-09-07
Add --mask-directive option

estum:
# Description

This PR adds the `--mask-directive` option similar to `--hide-tag` but for directives.

The following example will mask `@!parse` directive with `YARD::Tags::NilDirective` class to suppress any callbacks:

```
yardoc --mask-directive parse
```

Also, the option allows to suppress warnings on unknown directives in different contexts.

# Completed Tasks

### DIFF
--- a/lib/yard/cli/yardoc.rb
+++ b/lib/yard/cli/yardoc.rb
@@ -785,6 +785,12 @@ module YARD
         opts.on('--non-transitive-tag TAG', 'Marks a tag as not transitive') do |tag|
           Tags::Library.transitive_tags -= [tag.to_sym]
         end
+
+        opts.on('--mask-directive DIRECTIVE',
+                'Masks a directive to suppress any callback ' \
+                ' and the "unknown directive" warning') do |tag|
+          Tags::Library.define_directive(tag, Tags::NilDirective)
+        end
       end
     end
   end
--- a/lib/yard/tags/directives.rb
+++ b/lib/yard/tags/directives.rb
@@ -621,5 +621,11 @@ module YARD
         end
       end
     end
+
+    # Does nothing but suppresses "unknown directive" warnings, if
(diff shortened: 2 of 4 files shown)


estum:
Honestly, the second use-case was a main target for me and I explain why:

I have a project depended on `swagger-yard` gem  which defines several custom YARD tags & directives to declare docs for RESTful API in controllers by using YARD syntax. 

The same files may contain documentation for regular ruby methods, so I have tried to make a custom definitions to be ignored by yardoc. 

Whilst the `--hide-tag` option has already supplied to do it with custom tags, directives won't to be tamed and were keeping to annoy with `unknown directive` warnings. So, there is the dumbest solution I found.

lsegal:
Unfortunately this won't be incorporated into YARD. This seems better off like a situation where you could use plugins or locally loaded extensions to implement these adjustments rather than adding a global option to workaround what is very clearly a misconfiguration of some kind. This PR lacks good justification, use cases, and examples. Even if it were useful to you, I think there is limited benefit to other users and the benefit is not clearly explained.

Candidly, I'm not really understanding the use case.

> The same files may contain documentation for regular ruby methods.

I don't understand why this would be a problem. Directives should be usable alongside regular methods.

> Whilst the --hide-tag option has already supplied to do it with custom tags, directives won't to be tamed 

The premise of `--hide-tag` is to allow the use of data-only tags that do not show up in presentation, i.e. storing metadata like `@code_review.from_pull_request` that you don't necessarily need to see in generated docs (YARD will show these tags by default). Directives, on the other hand, are not themselves present in generated docs, because directives _are_ data-only.

tl;dr "directives" are not present in generated output and should not need to be hidden. If you are dealing with a directive that is generating unnecessary output, this is probably due to a misconfigured or poorly defined directive that needs fixing in upstream plugin code.

Issue #1230, opened 2019-03-13
XDG Base Directory Specification compliance


z0rc:
Yard uses `~/.yard` to store configuration, gem index, etc. It would be really nice to split this data to respectful folders defined by [XDG Base Dir Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Mainly config should go into `$XDG_CONFIG_HOME/yard/`, gem index and other generated data can be stored in `$XDG_CACHE_HOME/yard/`.

This would help with backups, keeping configuration separated from cache. But most importantly is that `$HOME` would be kept clear from unnecessary top level dotfiles. Check out how git, htop, systemd, mc and other CLI tools implement this spec.

lsegal:
I would accept support for XDG base directory support as a pull request, but I am going to close this issue out since we have a policy to not keep feature requests open.

If you want to take a look at implementing this I don't see any reason why we can't support it, as long as we fallback to `Dir.home` when these variables are not present.

soloturn:
@lsegal this seems more a bug than a feature request. the default should be the XDG default values not the $HOME/.yard, even when XDG_* environment variables are not set.

PR #1469, opened 2022-11-14
load project-specific configuration as well as user-specific

ccutrer:
so that you can configure a plugin per-project

# Completed Tasks

### DIFF
--- a/lib/yard/cli/yri.rb
+++ b/lib/yard/cli/yri.rb
@@ -5,14 +5,14 @@ module YARD
   module CLI
     # A tool to view documentation in the console like `ri`
     class YRI < Command
-      # The location in {YARD::CONFIG_DIR} where the YRI cache file is loaded
+      # The location in {YARD::USER_CONFIG_DIR} where the YRI cache file is loaded
       # from.
-      CACHE_FILE = File.expand_path(File.join(YARD::Config::CONFIG_DIR, 'yri_cache'))
+      CACHE_FILE = File.expand_path(File.join(YARD::Config::USER_CONFIG_DIR, 'yri_cache'))
 
       # A file containing all paths, delimited by newlines, to search for
       # yardoc databases.
       # @since 0.5.1
-      SEARCH_PATHS_FILE = File.expand_path(File.join(YARD::Config::CONFIG_DIR, 'yri_search_paths'))
+      SEARCH_PATHS_FILE = File.expand_path(File.join(YARD::Config::USER_CONFIG_DIR, 'yri_search_paths'))
 
       # Default search paths that should be loaded dynamically into YRI. These paths
       # take precedence over all other paths ({SEARCH_PATHS_FILE} and RubyGems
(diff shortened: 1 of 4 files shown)


lsegal (changes requested):
Added a few specific comments that would bar this from being accepted. In general I'm noticing quite a few breaking changes which are a blocker for merging. There are also a few deprecations, many of which do not seem necessary. It seems to me like project based configuration could be implemented in a completely additive fashion rather than changing any existing behaviors.

Overall, however, I think a little bit more thought needs to be put into the semantics of this feature for it to be valuable to various users / use cases. As it stands, it seems as though "project directory" is loosely defined by this PR as the current working directory, which is not exactly the most reliable way to do this. I think at the very least, this would need to be documented, but more accurately, I don't think you can really define a "project directory" without the ability to customize said directory. That makes this PR a little bit harder to build, as it would (ideally) require CLI options for users to configure / customize where their "project" is. Also, as a small sidenote here, I notice you only support the config file and not the full directory, which I think some users may have an issue with. Supporting this approach would lead to a bigger maintenance burden than implementing it correctly the first time around.

Another more generalized approach to this problem would be to simply go the `node_modules` route and search for a `.yard` directory in each parent directory to the root, followed by a `~/.yard` search. This would be compatible with eventually supporting #1230 as a final lookup location.

tl;dr simply implementing better lookup semantics for the `.yard` config directory would probably solve a lot of the needs for a project specific environment. Alternatively, simply having a CLI options that is also configurable via environment variable that points to said config dir / file / etc would probably solve 99% of the use cases here, even without directory lookup semantics.

ccutrer:
Ooh yay, YARD isn't abandoned!

I'm a little surprised at the pushback to the assumption that "project dir" == `.`. Isn't this how everything else in YARD works, right down to a .yardopts file? Is it even possible to run YARD anywhere except for from the project root? But that's okay... given a .yardopts file, I think it's reasonable to just add an explicit option to specify an explicit config dir.

That said, this is one of my bigger PRs, and it's been a while since I've worked on it. I have several others - most of them I have not even sent as PRs, since I just started pointing my project at my own fork of YARD. If the others go well, I'll start sending more PR, and then consider putting in the time addressing the comments on this one.

lsegal:
> I'm a little surprised at the pushback to the assumption that "project dir" == .

The issue isn't the assumption / default behavior, it's the lack of configurability. It's fine to assume the default, but YARD supports configuration of all of the other values, up to and including where the `.yardopts` file lives, where the `.yardoc` cache lives, and where any outputs go. The main issue here is that this PR has no such escape hatch / configuration.

ccutrer:
> > I'm a little surprised at the pushback to the assumption that "project dir" == .
> 
> The issue isn't the assumption / default behavior, it's the lack of configurability. It's fine to assume the default, but YARD supports configuration of all of the other values, up to and including where the `.yardopts` file lives, where the `.yardoc` cache lives, and where any outputs go. The main issue here is that this PR has no such escape hatch / configuration.

👍  ah okay, that makes more sense. I'll probably just re-implement as a switch to change the config dir, making sure that switch will be applied if it's in a yardopts file. That will be sufficient for my needs.

lsegal:
> [!NOTE]

Issue #1375, opened 2021-03-17
File traversal is recursive, which could result in SystemStackError


fsateler:
On a large project, it is easy to see lots of (debug) messages like this:

> [debug]: Missing object Item in file `app/forms/item/assign_form.rb', moving it to the back of the line.

Grepping for that phrase, yields this code in `ensure_loaded!`:

https://github.com/lsegal/yard/blob/359006641260eef1fe6d28f5c43c7c98d40f257d/lib/yard/handlers/base.rb#L566-L570

In particular, `parser.parse_remaining_files` causes a recursion:

![image](https://user-images.githubusercontent.com/1322013/111395246-285a6e00-869b-11eb-8a47-201a03950602.png)

On sufficiently large projects, this causes a SystemStackError.
exit.all

## Steps to reproduce

```
% seq 1 10000 | while read i ; do
echo "module Foo::My${i} ; end" > my_${i}.rb
done
% bundle exec yard doc --debug '*.rb'

```

## Actual Output

```
<snip>
[debug]: Parsing my_79.rb
[debug]: Missing object Foo in file `my_79.rb', moving it to the back of the line.
[debug]: Parsing my_80.rb
[debug]: Missing object Foo in file `my_80.rb', moving it to the back of the line.
[debug]: Parsing my_81.rb
[debug]: Missing object Foo in file `my_81.rb', moving it to the back of the line.
[debug]: Parsing my_82.rb
[debug]: Missing object Foo in file `my_82.rb', moving it to the back of the line.
[debug]: Parsing my_83.rb
<snip>
[debug]: Missing object Foo in file `my_556.rb', moving it to the back of the line.
[debug]: Parsing my_557.rb
[debug]: Missing object Foo in file `my_557.rb', moving it to the back of the line.
[debug]: Parsing my_558.rb
[debug]: Missing object Foo in file `my_558.rb', moving it to the back of the line.
[debug]: Parsing my_559.rb
[debug]: Missing object Foo in file `my_559.rb', moving it to the back of the line.
[debug]: Parsing my_560.rb
<snip>
bundler: failed to load command: yard (<home>/<gems>/ruby/2.6.0/bin/yard)
Traceback (most recent call last):
        10483: from <home>/.rbenv/versions/2.6.6/bin/bundle:23:in `<main>'
        10482: from <home>/.rbenv/versions/2.6.6/bin/bundle:23:in `load'
[...]

PR #1484, opened 2023-02-27
Fix stack overflow for large projects

Skipants:
# Description

When large projects have a lot of missing objects in the first pass of parsing Ruby files we would run into stack overflows. This happens because, when there was a missing object in a parsed Ruby file, we would recursively call `#parse_remaining_files`. When this happens a lot the stack would get huge.

We fix this by instead keeping a list of files that we want to retry and re-parse them in another pass. When we can no longer resolve any more files we break the loop.

Fixes https://github.com/lsegal/yard/issues/1375

# Completed Tasks


lib/yard/handlers/c/base.rb
```diff
@@ -75,20 +75,17 @@ def namespace_for_variable(var)
...
-            object = namespace_for_variable(var)
-            break unless object.is_a?(Proxy)
+          log.debug "Missing object #{object} in file `#{parser.file}', moving it to the back of the line."
 
-            raise NamespaceMissingError, object if retries > max_retries
-            log.debug "Missing namespace variable #{var} in file `#{parser.file}', moving it to the back of the line."
-            parser.parse_remaining_files
-            retries += 1
+          if globals.ordered_parser
+            retryable_file = parser.file == "(stdin)" ? StringIO.new("void Init_Foo() { #{statement.source} }") : parser.file
```
Skipants, line 84: Forgive my sin here. I wrote it this way because of tests that fail when we only try and use `parser.file` like in `lib/yard/handlers/base.rb`.

An example test that fails is https://github.com/lsegal/yard/blob/e91d41c032d1a61661ca34ecb2b67a2607bd875e/spec/handlers/c/class_handler_spec.rb#L70

Here's what happens/reasoning behind this weird hack:

1. Foo::Bar is not resolved and is instead retried in the first string of that test
2. Because this comes from stdin and not a file, `parser.file` is `(stdin)`.
3. If we push the string `(stdin)` onto the `globals.ordered_parser.files_to_retry` it's not able to be parsed by [OrderedParser#parse](https://github.com/Skipants/yard/blob/ec5be9e3bff08b6c49392b0b2002911be0fb62a9/lib/yard/parser/source_parser.rb#L41). It should instead be a `StringIO` with the C code contents.
4. The C code contents from `statement.source` are missing the wrapper `void Init_Foo() { ... }` and needs to be manually added.

An alternative to this was to save the contents in globals but that felt hacky as well for a couple reasons:

1. I was afraid pushing contents of files on globals could result in runaway mem usage (though now that I think about it, each new file probably clobbers the last one)
2. This is only useful in the niche case of C code from STDIN, so having a global for that felt like overkill and potentially confusing code

lsegal: This PR is definitely not mergeable with a hack like this, especially one that artificially causes a failing test to pass targeted specifically for that test.

Notably, (stdin) parsing is entirely common and not specific to C. What you're suggesting here is this PR does not support this use case. That would be a problem and highlights a possible breaking change.

lsegal (commented):
Sorry to get around to this so late. To be honest, the use of the specific workaround to C parsing is definitely concerning and makes me wonder if this will cause a breaking change, and thus I have not looked too deeply at merging. 

Unrolling the recursive loop might be useful here, but it would have to be done in a way that respects the existing order of operations and API capabilities. Supporting StringIOs is definitely one of those API capabilities.

I think this PR would need another pass to make it compatible.

lib/yard/handlers/c/base.rb
Skipants: Excellent feedback, thank you. I'll let this one simmer a bit and see if I can come up with a less-hacky patch.

lsegal:
> [!NOTE]

PR #1486, opened 2023-03-06
Preserve order of tags when reference tags are mixed with local tags

notEthan:
Currently, ref tags are always rendered after local tags, resulting, for me, in @param tags showing up in a different order than the actual method.
This PR removes separation of ref and local tags, so Docstring#tags can dereference and return tags in the right order.

# Completed Tasks


lsegal:
> [!NOTE]

PR #1487, opened 2023-03-06
Fix resolution of reference tags in a referenced docstring

notEthan:
When a docstring X is a reference to another docstring Y, any reference tags within docstring Y are resolved using the namespace of X, causing resolution to fail when X and Y are in different namespaces. This PR adds Docstring#namespace which follows the resolved reference.

# Completed Tasks

### DIFF
--- a/lib/yard/docstring.rb
+++ b/lib/yard/docstring.rb
@@ -46,6 +46,9 @@ module YARD
     # @return [CodeObjects::Base] the object that owns the docstring.
     attr_accessor :object
 
+    # @return [CodeObjects::Base] the namespace of the docstring
+    attr_reader :namespace
+
     # @return [Range] line range in the {#object}'s file where the docstring was parsed from
     attr_accessor :line_range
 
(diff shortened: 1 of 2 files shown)


lib/yard/docstring.rb
```diff
@@ -102,6 +106,7 @@ def self.new!(text, tags = [], object = nil, raw_data = nil, ref_object = nil)
     #   with.
     def initialize(content = '', object = nil)
       @object = object
+      @namespace = object
```
lsegal, line 109: It _kind of_ seems like `.namespace` is just a dupe of `.object`? Or, at the very least, that namespace could be a combination of `@object || @unresolved_reference` and thus no attribute would be needed.

lib/yard/docstring_parser.rb
```diff
@@ -224,7 +224,8 @@ def create_tag(tag_name, tag_buf = '')
 
     # Creates a {Tags::RefTag}
     def create_ref_tag(tag_name, name, object_name)
-      @tags << Tags::RefTagList.new(tag_name, P(object, object_name), name)
+      namespace = object && object.docstring ? object.docstring.namespace : object
+      @tags << Tags::RefTagList.new(tag_name, P(namespace, object_name), name)
```
lsegal, line 228: (See above) I could be wrong but it seems like `namespace` could be replaced entirely by `object || @unresolved_reference` and thus be the only necessary line of implementation.

lsegal (commented):
Just some thoughts.

lsegal:
> [!NOTE]

Issue #1362, opened 2020-11-23
Question: How to document methods that accept argument forwarding?


synthead:
Ruby 2.7 [introduces argument forwarding](https://bugs.ruby-lang.org/issues/16253), which looks like this:

```ruby
def accept_any_args(...)
  call_with_any_args(...)
end
```

If argument forwarding is used with YARD, one might attempt to document their methods like this:

```ruby
# Inspects some options.
#
# @param [Hash] options Options to inspect.
# @return [String] Inspected object.
def my_method(...)
  private_method(...)
end

private

def private_method(**options)
  options.inspect
end
```

However, when `yardoc` is ran with the above code, it raises this warning:

```
[warn]: @param tag has unknown parameter name: options 
```

How should the above code be documented with YARD?

lsegal:
The general approach here would be the same as the case where you have a `*args` or `**kwargs` method-- you would use `@overload` to define a synthesized method signature:

```ruby
# @overload my_method(a, b)
#   @param a [String] the first parameter
#   @param b [Boolean] the second parameter
def my_method(...) = other(...)
```

PR #1488, opened 2023-03-07
DocstringParser warnings improvements

notEthan:
Several improvements:
- Warnings about unknown params are skipped when the param is delegated to another method, determined by the param tag being a reference and the method taking a splat argument.
- param reference tags' names are used when possible.
- DocstringParser warnings include parser.object in the message - having the file/line number as it does now is sufficient, but jumping to the method name or class name can be simpler

The first one was what motivated this work, I wanted to get rid of warnings about delegated params. The rest were changes I made along the way.

# Completed Tasks


notEthan:
This is incomplete, it should handle argument forwarding with `def foo(...)` as well as splat. currently YARD::Handlers::Ruby::MethodHandler does not add a parameter for a `...` param, I'm having a look at what might be done with that. I could add a param named `...` there (haven't explored what else that might affect), but I am thinking maybe better to just drop the check for a splat/forwarding param here, not warn about unknown params that are (see ...) refs at all, trusting that they are handled.

also related: https://github.com/lsegal/yard/issues/1362

lib/yard/docstring_parser.rb
```diff
@@ -281,18 +281,33 @@ def self.after_parse_callbacks
...
+          if tag.name
+            param_tags << tag
+          elsif !(CodeObjects::Proxy === tag.owner)
+            param_tags += tag.tags
+          end
+          # if the ref tag doesn't have a name and doesn't yet resolve, do nothing
+        else
+          param_tags << tag
+        end
+      end
```
Skipants, line 298: You can flatten down this `if` and improve readability IMO by using `.filter_map` to iterate here. Consider:
```ruby
param_tags = parser.tags.filter_map do |tag|
  next unless tag.tag_name == "param"
  next tag if !tag.is_a?(Tags::RefTagList) || tag.name
  
  tag.tags unless CodeObjects::Proxy === tag.owner
end
```

`filter_map` is Ruby 2.7+ and I'm not sure about Yard's current minimum Ruby version, so if it's lower you can use `map{ ... }.compact` instead.

Skipants (commented):
I hope you don't mind some random reviewing your PR. I'm using Yard a lot lately and am invested in helping out.

lsegal:
> [!NOTE]

PR #1489, opened 2023-03-07
github CI use macos-latest

notEthan:
macos-10.15 is unsupported, see https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/
This could be changed to macos-12 or macos-11 if there is reason to reference a particular version, but using -latest seemed to me like a good option.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


notEthan:
Maybe also switching to ubuntu-latest and windows-latest for consistency would be good, or maybe I should not use any -latest and use macos-12.

lsegal:
Thanks for the PR!

PR #1510, opened 2023-09-24
Parser::SourceParser#convert_encoding - fixup BOM encoding

### LINKED ISSUE #1509: ruby3.3 causes failure for YARD::Parser::SourceParser#parse understands UTF-8 BOM
With ruby 3.3 (I tested https://github.com/ruby/ruby/commit/2ceb5363f9b15cffd3901220463a1175a32b89f7 ) yard 0.9.34 ( https://github.com/lsegal/yard/commit/698b22cc2340230856ab17a25cfefa889f1c8289 ) spec testsuite fails like:

```
$ rspec -r spec_helper spec

Failures:

  1) YARD::Parser::SourceParser#parse understands UTF-8 BOM
     Failure/Error: expect(Registry.all(:class).first.path).to eq "FooBar"

       expected: "FooBar"
            got: "FooBa"

       (compared using ==)
     # ./spec/parser/source_parser_spec.rb:589:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:129:in `block (2 levels) in <top (required)>'
```

ruby 3.2.2 does not see this error. For now I don't know what ruby side change caused this.

MSP-Greg:
# Description

Ruby 3.3 may change handling of BOM encoded files.  Change code to work with 3.3 and earlier versions.

Closes #1509 

Note that Ruby 2.7 added [`IO##set_encoding_by_bom`](https://msp-greg.github.io/ruby_2_7/Core/IO.html#set_encoding_by_bom-instance_method)

# Completed Tasks

### DIFF
--- a/lib/yard/parser/source_parser.rb
+++ b/lib/yard/parser/source_parser.rb
@@ -476,9 +476,8 @@ module YARD
           content.force_encoding('binary')
           ENCODING_BYTE_ORDER_MARKS.each do |encoding, bom|
             bom.force_encoding('binary')
-            if content[0, bom.size] == bom
-              content.force_encoding(encoding)
-              return content
+            if content.start_with?(bom)
+              return content.sub(bom, '').force_encoding(encoding)
             end
           end
           content.force_encoding('utf-8') # UTF-8 is default encoding


lib/yard/parser/source_parser.rb
```diff
@@ -476,9 +476,8 @@ def convert_encoding(content)
           content.force_encoding('binary')
           ENCODING_BYTE_ORDER_MARKS.each do |encoding, bom|
             bom.force_encoding('binary')
-            if content[0, bom.size] == bom
-              content.force_encoding(encoding)
-              return content
+            if content.start_with?(bom)
```
lsegal, line 479: Not sure there's a good enough reason to switch to this more restrictive (newer) API given the simplicity of the existing line.

lsegal (changes requested):
I think this would be good if it retained the initial equality check.

lib/yard/parser/source_parser.rb
MSP-Greg: I changed it for two reasons, first, `content[0, bom.size]` creates a string, second, both `content` and `bom` are both forced to `binary` encoding, so there won't be any encoding issues with a character based test, which also happens with the existing check.  Not sure why you're calling it 'more restrictive'?

Regardless, I can revert that change if you'd like...

lsegal: I remember start_with? being added in Ruby 2.5+, but it seems like it was added much earlier-- in which case this is fine.

lsegal (approved):

lsegal:
Thanks for the quick fix on #1509!

lib/yard/parser/source_parser.rb
MSP-Greg: > I remember start_with? being added in Ruby 2.5+

My memory isn't good for that sort of thing.  CI did pass...  I remember a `starts_with?` a very long time ago.  Thanks.

PR #1516, opened 2023-10-09
Add doc for Tag#name if no tag name is supplied

ksss:
# Description

Added documentation for `YARD::Tags::Tag#name`, which could be `nil`.

The following is a typical example.

```rb
YARD::DocstringParser.new.parse("@return [void]").tags
#=> [#<YARD::Tags::Tag:0x0000000105294108 @name=nil, @tag_name="return", @text="", @types=["void"]>]
```

# Completed Tasks


lsegal:
Makes sense, thanks!

PR #1535, opened 2024-02-26
Don't update window location in frames with javascript

alextwoods:
# Description

Don't update the window location in frames.erb when the input contains javascript.

# Completed Tasks


RedYetiDev:
Hi! This pull request won't resolve the issue. XSS will still be possible with `data:` URLs

alextwoods:
Good point - I hadn't considered `data:` since I thought most modern browsers block them in the top frame, but should likely be handled as well.

RedYetiDev:
I actually currently have an advisory open to resolve this issue, if you'd like I can show you my proposed patch

PR #1538, opened 2024-02-29
Update frames.erb for better XSS check

RedYetiDev:
# Description

Better XSS detection using `URL`.

# Completed Tasks

### DIFF
--- a/templates/default/fulldoc/html/frames.erb
+++ b/templates/default/fulldoc/html/frames.erb
@@ -5,7 +5,7 @@
 	<title><%= options.title %></title>
 </head>
 <script type="text/javascript">
-var mainUrl = '<%= url_for_main %>'
+var mainUrl = '<%= url_for_main %>';
 try {
     var match = decodeURIComponent(window.location.hash).match(/^#!(.+)/);
     var name = match ? match[1] : mainUrl;
--- a/templates/default/fulldoc/html/frames.erb
+++ b/templates/default/fulldoc/html/frames.erb
@@ -5,10 +5,14 @@
 	<title><%= options.title %></title>
 </head>
 <script type="text/javascript">
-  var match = decodeURIComponent(window.location.hash).match(/^#!(.+)/);
-  var name = match ? match[1] : '<%= url_for_main %>';
-  name = name.replace(/^((\w*):)?[\/\\]*/gm, '').trim();
-  window.top.location.replace(name)
+try {
+    var match = decodeURIComponent(window.location.hash).match(/^#!(.+)/);
+    var name = match ? match[1] : '<%= url_for_main %>';
+    var url = new URL(name, location.href);
+    window.top.location.replace(url.origin === location.origin ? name : '<%= url_for_main %>');
+} catch (e) {
+    window.top.location.replace('<%= url_for_main %>');
+}
 </script>
 <noscript>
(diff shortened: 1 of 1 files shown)


templates/default/fulldoc/html/frames.erb
```diff
@@ -5,10 +5,14 @@
...
 <script type="text/javascript">
-  var match = decodeURIComponent(window.location.hash).match(/^#!(.+)/);
-  var name = match ? match[1] : '<%= url_for_main %>';
-  name = name.replace(/^((\w*):)?[\/\\]*/gm, '').trim();
-  window.top.location.replace(name)
+try {
+    var match = decodeURIComponent(window.location.hash).match(/^#!(.+)/);
+    var name = match ? match[1] : '<%= url_for_main %>';
+    var url = new URL(name, location.href);
+    window.top.location.replace(url.origin === location.origin ? name : '<%= url_for_main %>');
```
lsegal, line 12: at this point you might want to assign `'<%= url_for_main %>'` to a variable.

RedYetiDev: I don't think its needed for that. While it is commonly referenced, it's not stored in the browser memory until it is needed. I think this, while a bit uglier, is more browser-efficient.

lsegal: It's less efficient on the generation side since we're unnecessarily calling `url_for_main` 2 extra times. Also from a DRY perspective, it we should only have the one variable.

RedYetiDev: Okay, would you like to me to update the code, or will you?

templates/default/fulldoc/html/frames.erb
```diff
@@ -5,10 +5,14 @@
...
+} catch (e) {
+    window.top.location.replace('<%= url_for_main %>');
+}
```
lsegal:
```suggestion
  try {
    var mainUrl = '<%= url_for_main %>';
    var match = decodeURIComponent(window.location.hash).match(/^#!(.+)/);
    var name = match ? match[1] : mainUrl;
    var url = new URL(name, location.href);
    window.top.location.replace(url.origin === location.origin ? name : mainUrl);
  } catch (e) {
    window.top.location.replace(mainUrl);
  }
```

RedYetiDev: The assignment must be outside the try catch loop

templates/default/fulldoc/html/frames.erb
```diff
@@ -5,10 +5,15 @@
 	<title><%= options.title %></title>
 </head>
 <script type="text/javascript">
-  var match = decodeURIComponent(window.location.hash).match(/^#!(.+)/);
-  var name = match ? match[1] : '<%= url_for_main %>';
-  name = name.replace(/^((\w*):)?[\/\\]*/gm, '').trim();
-  window.top.location.replace(name)
+var mainUrl = '<%= url_for_main %>'
```
lsegal, line 8:
```suggestion
var mainUrl = '<%= url_for_main %>';
```

lsegal:
LGTM

RedYetiDev:
Same here

RedYetiDev:
Great! I'm sorry again for my initial mistake and I'm glad we worked it out!

PR #1540, opened 2024-03-23
Commonmarker 1.0 support

### LINKED ISSUE #1528: Commonmarker 1.0 support
Commonmarker 1.0 changes the top-level constant name from `CommonMarker` to `Commonmarker`, so YARD fails to detect the provider.

https://github.com/lsegal/yard/blob/2d197a381c5d4cc5c55b2c60fff992b31c986361/lib/yard/templates/helpers/markup_helper.rb#L33

https://github.com/gjtorikian/commonmarker/issues/269

ParadoxV5:
# Description

* Add `Commonmarker` (all lower-case `m`s) provider for `commonmarker ~> 1.0.0`
  * Fixes #1528
* Also changes “Missing markup renderer” from `skip` to `raise` to catch this sort of issues in the future

# Completed Tasks

### DIFF
--- a/lib/yard/templates/helpers/html_helper.rb
+++ b/lib/yard/templates/helpers/html_helper.rb
@@ -89,6 +89,8 @@ module YARD
                              :tables,
                              :with_toc_data,
                              :no_intraemphasis).to_html
+        when 'Commonmarker'
+          Commonmarker.to_html(text) # GFM configs are on by default
         when 'CommonMarker'
           CommonMarker.render_html(text, %i[DEFAULT GITHUB_PRE_LANG], %i[autolink table])
         else
--- a/lib/yard/templates/helpers/html_helper.rb
+++ b/lib/yard/templates/helpers/html_helper.rb
@@ -90,7 +90,8 @@ module YARD
                              :with_toc_data,
                              :no_intraemphasis).to_html
         when 'Commonmarker'
-          Commonmarker.to_html(text) # GFM configs are on by default
+          # GFM configs are on by default; use YARD for syntax highlighting
+          Commonmarker.to_html(text, plugins: {syntax_highlighter: nil})
         when 'CommonMarker'
           CommonMarker.render_html(text, %i[DEFAULT GITHUB_PRE_LANG], %i[autolink table])
         else
--- a/lib/yard/templates/helpers/markup_helper.rb
+++ b/lib/yard/templates/helpers/markup_helper.rb
@@ -30,6 +30,7 @@ module YARD
           {:lib => :maruku, :const => 'Maruku'},
           {:lib => :'rpeg-markdown', :const => 'PEGMarkdown'},
           {:lib => :rdoc, :const => 'YARD::Templates::Helpers::Markup::RDocMarkdown'},
+          {:lib => :commonmarker, :const => 'Commonmarker'},
           {:lib => :commonmarker, :const => 'CommonMarker'}
         ],
         :textile => [
--- a/spec/templates/markup_processor_integrations/integration_spec_helper.rb
+++ b/spec/templates/markup_processor_integrations/integration_spec_helper.rb
@@ -26,7 +26,7 @@ RSpec.shared_context 'shared helpers for markup processor integration specs' do
 
   before(:each) do
     if html_renderer.markup_class(markup).nil?
-      skip "Missing markup renderer #{markup}"
+      raise "Missing markup renderer #{markup}"
     end
   end
 
--- a/spec/templates/markup_processor_integrations/markdown_spec.rb
+++ b/spec/templates/markup_processor_integrations/markdown_spec.rb
@@ -82,7 +82,7 @@ MARKDOWN
     end
   end
 
-  describe 'CommonMarker', if:  RUBY_VERSION >= '2.3' do
+  describe 'Commonmarker', if:  RUBY_VERSION >= '2.3' do
     let(:markup) { :markdown }
     let(:markup_provider) { :commonmarker }
 
--- a/spec/templates/helpers/html_helper_spec.rb
+++ b/spec/templates/helpers/html_helper_spec.rb
@@ -158,7 +158,7 @@ RSpec.describe YARD::Templates::Helpers::HtmlHelper do
 
     it "creates tables (markdown specific)" do
       log.enter_level(Logger::FATAL) do
-        supports_table = %w(RedcarpetCompat Kramdown::Document CommonMarker)
+        supports_table = %w(RedcarpetCompat Kramdown::Document Commonmarker CommonMarker)
         unless supports_table.include?(markup_class(:markdown).to_s)
           pending "This test depends on a markdown engine that supports tables"
         end


lib/yard/templates/helpers/html_helper.rb
```diff
@@ -89,6 +89,9 @@ def html_markup_markdown(text)
                              :tables,
                              :with_toc_data,
                              :no_intraemphasis).to_html
+        when 'Commonmarker'
+          # GFM configs are on by default; use YARD for syntax highlighting
+          Commonmarker.to_html(text, plugins: {syntax_highlighter: nil})
```
ParadoxV5, line 94:
```suggestion
          Commonmarker.to_html(text)
```
However, preference for the provider’s highlighting is outside of this PR’s scope.

spec/templates/markup_processor_integrations/integration_spec_helper.rb
```diff
@@ -26,7 +26,7 @@
 
   before(:each) do
     if html_renderer.markup_class(markup).nil?
-      skip "Missing markup renderer #{markup}"
+      raise "Missing markup renderer #{markup}"
```
lsegal, line 29: This should stay as skip for systems that do not have the dependencies installed (testing outside of bundle exec).

lib/yard/templates/helpers/markup_helper.rb
```diff
@@ -30,6 +30,7 @@ def clear_markup_cache
           {:lib => :maruku, :const => 'Maruku'},
           {:lib => :'rpeg-markdown', :const => 'PEGMarkdown'},
           {:lib => :rdoc, :const => 'YARD::Templates::Helpers::Markup::RDocMarkdown'},
+          {:lib => :commonmarker, :const => 'Commonmarker'},
```
lsegal, line 33: This is problematic, as the `lib` value is a unique key passed via `-m`. This will effectively mean that the `CommonMarker` (pre 1.0?) class could never be activated. We should definitely maintain support for the old format, which would require this commonmarker key be changed-- it's not idela, but maybe `:commonmarker1`?

spec/templates/markup_processor_integrations/markdown_spec.rb
```diff
@@ -72,26 +72,17 @@
...
   end
 
-  describe 'CommonMarker', if:  RUBY_VERSION >= '2.3' do
+  describe 'Commonmarker', if: RUBY_VERSION >= '2.3' do
     let(:markup) { :markdown }
     let(:markup_provider) { :commonmarker }
 
     include_examples 'shared examples for markdown processors'
 
-    it 'generates level 2 header without id' do
```
lsegal, line 91: This should remain but should be scoped to CommonMarker

lsegal (changes requested):
There's an issue in supporting both CommonMarker and Commonmarker in this PR. At first glance it seems like YARD is no longer able to support the pre 1.0 commonmarker lib due to a change in the markup_helper lookup table. If we can address that, I think this mostly looks good.

lib/yard/templates/helpers/markup_helper.rb
ParadoxV5: `CommonMarker` seems still accessible.
https://github.com/lsegal/yard/blob/ad2c1c4d7c53f71649440c816104ef0077226cb2/lib/yard/templates/helpers/markup_helper.rb#L91-L110

---

Even if it’s not, v1.0 `CommonMaker` should have priority over pre-1.0 `Commonmarker`.
Separating a `commonmarker1` key keeps `commonmarker` out of date.
YARD does not have a system that checks provider versions and has long[^t] exposed users to Commonmarker 1.0’s backward incompatibility. We should improve this compatibility rather than make users work an extra step.

[^t]: [Commonmarker 1.0 released](https://github.com/gjtorikian/commonmarker/releases/tag/v1.0.0) last Christmas Eve!

spec/templates/markup_processor_integrations/markdown_spec.rb
ParadoxV5: Why? Does it have an impact?

* > Testing providers’ implementation is outside of the scope of YARD.
* > Remove tests for Markdown header IDs or the lack of, considering they may improve with newer provider versions and that YARD shouldn’t constraint their growth

lsegal: > Testing providers’ implementation is outside of the scope of YARD.

Not when we're explicitly testing system integration. The fact that this behavior changes is exactly why we have tests. An id-less h2 heading actually has implications for downstream YARD code, since we have special code paths for heading id attributes.

Actually, expanding this comment outward, it seems like this change removes all integration tests for CommonMarker, which means we're now only testing Commonmarker 1.0. This is also insufficient. We should keep the old tests for CommonMarker and add the new ones for this 1.0 version, not replace.

tl;dr we need to test both CommonMarker / Commonmarker implementations. Not just one. You can remove the h2 test for Commonmarker 1.0 if you want, but both libraries should be integration tested.

lsegal: > Remove tests for Markdown header IDs or the lack of, considering they may improve with newer provider versions and that YARD shouldn’t constraint their growth

YARD relies on this IDs though, that's the issue, and that's why we test.

lib/yard/templates/helpers/markup_helper.rb
lsegal: > Even if it’s not, v1.0 CommonMaker should have priority over pre-1.0 Commonmarker.

Unfortunately this would be a breaking change and would need a major version bump-- we're very unlikely to major version bump YARD just for a single (optional) markup library.

ParadoxV5: YARD is currently still in 0.x versions, so I’m not sure what you meant by ‘major’.

That said, together with https://github.com/lsegal/yard/pull/1540#discussion_r1731826662, I see your concern.
YARD currently has designs tailored for CommonMarker v0.x but it’s not documented anywhere.
We can only treat Commonmarker v1.x as a separate provider.

ParadoxV5: > YARD does not have a system that checks provider versions

We might wanna add one. But for now, …

> YARD currently has designs tailored for CommonMarker v0.x but it’s not documented anywhere.

… I’ll add a comment in the code since it’s now linked in the README.
* Prerequisite: #1544

ParadoxV5: > At first glance it seems like YARD is no longer able to support the pre 1.0 commonmarker lib due to a change in the markup_helper lookup table.

Actually, we might not be able to support both Commonmarker and CommomMarker at all.
Commonmarker 1.x shadows CommonMarker 0.x. We’d have to write individual `Bundler.require`s in our specs.

spec/templates/markup_processor_integrations/integration_spec_helper.rb
ParadoxV5: I am unable to run `rake`/`rspec` locally without a prior `bundle install`.

noraj:
I'm not sure if I understood all that was said here.

Why support several versions of commonmarker, why isn't yard supporting only the last version for each markup? Yes that's a potential breaking (of an optional feature, not of yard core), introducing new major version, and so what? People that want to keep older versions of the markup can still use older version of yard. Else, yard will still always be locked using outdated versions of markups. That would also be less cumbersome and maintenance work than having a system for version requirements and maintaining several plugins for each major version of each markup. If only one version of commonmarker should be supported, it should be the latest.

lsegal:
Closing as stale. If we want to improve support for CommonMarker it has to be done in a backward compatible way. An easy way to do this would be to introduce a back compat layer that can handle either version. Unfortunately yard can't break behavior with old gem versions.

noraj:
https://github.com/lsegal/yard/pull/1663#issuecomment-4230094749

> Feedback was already provided on https://github.com/lsegal/yard/pull/1540 that was never addressed. That PR should probably be closed as stale. It would be easy to open a PR that addresses the feedback by creating a proper compatibility layer and does not break backward compatibility. You're free to open such a PR.

I'd be glad to open such a PR. But I fear what seems easy for you, is not for me. I'm not sure I understood why this PR can't be merged.

In this comment https://github.com/lsegal/yard/pull/1540/changes#r1730792625, you said the new code is problematic because `:lib` should be a unique value.

```diff
-{:lib => :commonmarker, :const => 'Commonmarker'},
+{:lib => :commonmarker1, :const => 'Commonmarker'},
{:lib => :commonmarker, :const => 'CommonMarker'}
```

So using `:commonmarker1` for CM 1.x / 2.x would be enough?

For the spec, you said it should stay as skip but paradox say it still requires the gem? You have not yet answered to that https://github.com/lsegal/yard/pull/1540/changes#r1730786139.

https://github.com/lsegal/yard/pull/1540/changes#r1730794283 For the tests, what should we do? Let CommonMarker tests as it and duplicate them for Commonmarker?

Is there anything else needed? You were talking about a compatibility layer? Do you meant that instead of having `:commonmarker` and `:commonmarker1` we should only have `:commonmarker` and have `:const => commonmarkerDectection` where `commonmarkerDectection` would be a method returning Commonmarker or CommonMarker based on the installed version installed?

noraj:
Please tell me what to do before I spend time working on a PR.

ParadoxV5:
Hi @noraj, there’s another PR from the maintainer of Commonmarker themself, #1601.

PR #1541, opened 2024-03-23
Support Ruby 3 inline attribute visibility

theodorton:
See https://redmine.ruby-lang.org/issues/17314

# Description

Currently attributes defined with:

```rb
private attr_accessor :foo
protected attr_accessor :bar
```

Will be displayed as public attributes in the documentation.

These should be considered private/protected and not be included in the documentation.

When generating documentation, this warning is show:

```
[warn]: in YARD::Handlers::Ruby::VisibilityHandler: Undocumentable statement, cannot determine method name
	in file ...:

	1: private attr_accessor :foo
```

# Completed Tasks

### DIFF
--- a/lib/yard/handlers/ruby/visibility_handler.rb
+++ b/lib/yard/handlers/ruby/visibility_handler.rb
@@ -13,10 +13,22 @@ class YARD::Handlers::Ruby::VisibilityHandler < YARD::Handlers::Ruby::Base
     case statement.type
     when :var_ref, :vcall
       self.visibility = ident.first.to_sym
-    when :fcall, :command
+    when :command
+      if RUBY_VERSION >= '3.' && is_attribute_method?(statement.parameters.first)
+        parse_block(statement.parameters.first, visibility: ident.first.to_sym)
+        return
+      end
+      process_decorator do |method|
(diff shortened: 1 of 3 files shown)


lib/yard/handlers/ruby/visibility_handler.rb
```diff
@@ -13,10 +13,22 @@ class YARD::Handlers::Ruby::VisibilityHandler < YARD::Handlers::Ruby::Base
     case statement.type
     when :var_ref, :vcall
       self.visibility = ident.first.to_sym
-    when :fcall, :command
+    when :command
+      if RUBY_VERSION >= '3.' && is_attribute_method?(statement.parameters.first)
```
lsegal, line 17: Ruby version check seems unnecessary since (a) this will check the version of Ruby parsing the source file, not the version that the source file was intended for and (b) since (a) is true, as long as ripper parses correctly, we should assume the code was written for the latest Ruby (we don't know what the target Ruby the .rb file was written for).

```suggestion
      if is_attribute_method?(statement.parameters.first)
```

lsegal (approved):
Looks good

theodorton:
@lsegal I can look into the failing specs here and request a second review.

PR #1588, opened 2024-09-28
Update Templates.md to fix incorrect path structure

topherfangio:
# Description

Fix incorrect path structure in "Overriding Templates by Registering a Template Path" section.

Because of the `YARD::Templates::Engine.register_template_path '/path/to/mytemplates'` code above, we need a `default` subdirectory to contain the `class/html/customsection.erb` and `class/setup.rb` files.

The paragraph immediately after this even mentions the `default/class` structure, but I missed it since it wasn't in the tree structure.

Tested with yard 0.9.36 to verify.

# Completed Tasks

### DIFF
--- a/docs/Templates.md
+++ b/docs/Templates.md
@@ -280,10 +280,11 @@ seen above by creating such a path in our '/path/to/mytemplates' custom template
 path:
 
     /path/to/mytemplates/:
-    |-- class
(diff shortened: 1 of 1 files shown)


lsegal:
Thanks for the fix!

PR #1590, opened 2024-10-11
Update signature handling in @overload

pvande:
# Description

Previously, `@overload` tags that defined required keyword arguments would end up with an empty default value, to the delight of nobody. While this might *ideally* be handled by moving away from the legacy parsers, this is a stopgap that makes an already uncommon edge case go away.

# Completed Tasks

### DIFF
--- a/spec/tags/overload_tag_spec.rb
+++ b/spec/tags/overload_tag_spec.rb
@@ -50,4 +50,19 @@ RSpec.describe YARD::Tags::OverloadTag do
     tag = Tags::OverloadTag.new(:overload, "default")
     expect(tag.signature).to eq "default"
   end
+
+  it "properly handles complex signatures" do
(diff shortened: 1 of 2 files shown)


lsegal (approved):
Makes sense and looks good.

lsegal:
Thanks for this fix!

spec/tags/overload_tag_spec.rb
```diff
@@ -50,4 +50,19 @@ def bar(a, b = 1, &block)
     tag = Tags::OverloadTag.new(:overload, "default")
     expect(tag.signature).to eq "default"
```
carveR1209, line 51: -1(/+37👀**
```suggestion
     expect(tag.signature).to eq "default"-1)+31🚥
```
**@carveR1209@carveR1209🚥

carveR1209: @github/@carveR1209🛜

PR #1598, opened 2024-12-20
Ensure YardocTask requires yard itself

### LINKED ISSUE #1597: Yard rake task is missing require
## Steps to reproduce

The yard rake task is not self-contained and is missing requires to yard itself.

1. add yard to rakefile:
    ```rb
    require "yard/rake/yardoc_task"
    YARD::Rake::YardocTask.new
    ````
2. verify yard rake task:
    ```
    $ rake -T
    # snip
    rake yard                     # Generate YARD Documentation
    ```
3. run `rake yard` task

## Actual Output

    ```
    $ rake yard
    rake aborted!
    NameError: uninitialized constant YARD::CLI

              yardoc = YARD::CLI::Yardoc.new
                           ^^^^^

    Tasks: TOP => yard
    (See full trace by running task with --trace)
    ```

## Expected Output

I expected roughly the same output as `yard`

It seems the yard task is missing requires.

## Environment details:

- OS: macos 15.2
- Ruby version (`ruby -v`): ruby 3.1.6p260 (2024-05-29 revision a777087be6) [arm64-darwin23]
- YARD version (`yard -v`): yard 0.9.37
- Relevant software dependency/versions: rake, version 13.2.1

jasonkarns:
# Description

The yard task is not self contained at the moment and does not run without yard itself first being loaded.

It's preferable for rake tasks to be self-contained (the YardocTask should require its dependencies so users may require just the tasklib itself).

Fixes #1597 

# Completed Tasks

  - [x] `rake spec` passes locally. No tests added yet. Wanting feedback first before spending time deciphering test suite.

### DIFF
--- a/lib/yard/rake/yardoc_task.rb
+++ b/lib/yard/rake/yardoc_task.rb
@@ -68,6 +68,9 @@ module YARD
       def define
         desc "Generate YARD Documentation" unless ::Rake.application.last_description
         task(name) do
+          # Lazy load yard so we don't impact the load time of the Rakefile
+          require_relative '../../yard'
+
           before.call if before.is_a?(Proc)
           yardoc = YARD::CLI::Yardoc.new
           yardoc.options[:verifier] = verifier if verifier


lib/yard/rake/yardoc_task.rb
```diff
@@ -68,6 +68,8 @@ def initialize(name = :yard)
       def define
         desc "Generate YARD Documentation" unless ::Rake.application.last_description
         task(name) do
+          require File.expand_path('../../yard', __dir__)
```
jasonkarns, line 71: Would prefer simple `require_relative` if supported rubies allow for it.

jasonkarns: switched to require_relative since it's available since ruby 1.9.2.

lsegal:
Based on #1597, the intended API for loading the YardocTask is to `require 'yard'` so as to not introduce an implicit potential breaking change dependency on the file system structure of the YARD codebase. As such, this change should generally not be necessary. The majority of YARD is loaded lazily (except config loading, which is roughly a no-op if a config is not present), so unless there is specific data that shows a performance issue with this approach, I think it's best to stick with the safer API.

PR #1600, opened 2025-01-17
Support A = Data.define(:a, :b, :c)

marshall-lee:
# Description

`Data` was added in Ruby 3.2. This is similar to `Struct` but immutable and with kwargs by default. So this PR aims to support simple `Data.define` const assignments:

```ruby
Foo = Data.define(:a, :b, c)
Foo::Bar = Data.define(:a, :b, c)
Baz = Foo = Data.define(:a, :b, c) do
  def d; end
end
```

Note: the support of the `class Foo < Data.define(:a, :b)` form was not added by intention because I don't want to encourage this style of programming and [it does not appear in the `Data` documentation](https://docs.ruby-lang.org/en/3.2/Data.html). If one has a strong opinion that this style must be supported, they're free to do it in a separate PR.

# Completed Tasks


lsegal:
Great change, thank you!

PR #1604, opened 2025-02-01
Fix array-within-word-array parsing

apiology:
# Description

When yard saw a constant like:

```ruby
FOO = [%w[bar baz]]
```

it cut off the array at the first closing bracket in the resulting documentation:

```ruby
[%w[bar baz]
```

Fix by ensuring that these types of arrays, which don't have a
starting :lbracket, don't pop off :lbracket state in `@map`

Drop dead code and add a variety of related regression cases found in
my application using sord after trying different fixes.

# Completed Tasks

- [X] I wrote tests and ran `bundle exec rake` locally (if code is attached to PR).

### DIFF
--- a/lib/yard/parser/ruby/ruby_parser.rb
+++ b/lib/yard/parser/ruby/ruby_parser.rb
@@ -390,7 +390,7 @@ module YARD
 
         def on_array(other)
           node = AstNode.node_class_for(:array).new(:array, [other])
-          map = @map[MAPPINGS[node.type]]
+          map = @map[MAPPINGS[node.type]] if other.nil? || other.type == :list
           if map && !map.empty?
             lstart, sstart = *map.pop
             node.source_range = Range.new(sstart, @ns_charno - 1)
--- a/spec/parser/ruby/ruby_parser_spec.rb
+++ b/spec/parser/ruby/ruby_parser_spec.rb
@@ -283,6 +283,11 @@ eof
         src = "%#{tok}{\na b c\n d e f\n}"
         expect(stmt(src).source).to eq src
       end
+
+      it "shows proper source for %#{tok}[] array" do
+        src = "%#{tok}[\na b c\n d e f\n]"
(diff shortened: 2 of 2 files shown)


apiology:
Hmmm, it looks like `['1', %w[]]` fails as well.

apiology:
All right, added simplified regression test cases, found a different fix which seems more robust, and verified it passed tests as well as sord on my application.

sord is useful for regression against a codebase here as it uses the output of yard's parse of e.g. constants and feeds the results into sorbet, which parses it again, complaining about syntax errors when yard doesn't give the correct source range.

lib/yard/parser/ruby/ruby_parser.rb
```diff
@@ -390,16 +390,12 @@ def on_aref_field(*args)
...
+          map = @map[MAPPINGS[node.type]] if other.nil? || other.type == :list
           if map && !map.empty?
             lstart, sstart = *map.pop
             node.source_range = Range.new(sstart, @ns_charno - 1)
             node.line_range = Range.new(lstart, lineno)
           else
-            sstart = other.source_range.begin
-            lstart = other.line_range.begin
-            node.source_range = Range.new(sstart, @ns_charno - 1)
-            node.line_range = Range.new(lstart, lineno)
```
apiology, line 402: Unless I'm missing something, this is dead code.  Let me know if I'm wrong.

lsegal: This is not dead code per se. This is meant to future-proof the implementation for new node types which happens fairly often.

apiology: Gotcha.  Just seemed strange those two assignments were immediately overridden.  I'll revert that, but definitely curious to hear what I'm misunderstanding there.

apiology: Reverted

lsegal:
Thanks for the fix!

PR #1607, opened 2025-02-25
Remove rackup server

xjunior:
# Description

`Rackup::Server` was removed on 1.0.1. There's no much explanation on [their changelog](https://github.com/rack/rackup/blob/main/releases.md), and the git blame doesn't give up much information either. Fact is that this class doesn't exist.

The first commit simply adds this dependency, which breaks the specs. The second fixes it.

# Completed Tasks

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,7 @@ end
 
 group :server do
   gem 'rack', '~> 2.0'
+  gem 'rackup', '~> 1.0'
 end
 
 group :i18n do


wadewinningham (approved):

xjunior:
@lsegal `Rackup::Server` was removed on `rackup` 1.0.1.  So requiring `rackup` works, but the constant doesn't exist and this breaks. This started happening for us on Rails 7.1, as it adds a direct dependency to rackup >~ 1.0

xjunior:
@lsegal if you go back to the first commit of this branch and run the specs you'll see them fail

xjunior:
@lsegal it's really confusing, honestly. If you look into rackup the change from 1.0.0 to 1.0.1 is [`Fix rackup.rb invalid requires.`](https://github.com/rack/rackup/blob/main/releases.md). But in fact they removed everything from the project. Since there's no proper history on this change, I'm not sure which one is broken, but it's a fact that YARD is affected by this in Rails 7.1.

lsegal:
Seems like a bug that should be reported to the rackup project as they broke compatibility across versions.

xjunior:
@lsegal what concerns me is the update message "remove invalid require". There's no much info about what they meant with that, and whether that's actually accurate, as what they did in fact was wipe up the library. Assuming wiping was wrong, this lib would have to explicitly require rackup/server instead.

Regardless, I'm making this contribution as a supporter and user of both libraries, but in our case we're just removing the Yard::Server we have hosted for our internal APIs as this is not our primary way of using them. That unblocks us.

lsegal:
> what concerns me is the update message "remove invalid require". 

It must be a mistake, as rackup 2.x retains this constant, but I will leave it up to them to respond properly.

PR #1608, opened 2025-03-18
refactor RDoc markup handling for improved loading and initialization

seuros:
# Description

rdoc class had support for a rdoc versin that is 12 years old.

added https for the homepage.

# Completed Tasks


lsegal:
This seems to fail all tests. Can you look into this?

seuros:
Oh cool, i see you dropped legacy support for 2.6.

Do you think we should go at a safer zone to support only 3.2+ since anything before is EOL.

I have some yard plugins that work only in those version as they use edge gem or ruby features.

lsegal:
Actually, the tests are being skipped because the suite isn't configured to work with 1.0+.

https://github.com/lsegal/yard/actions/runs/14940811833/job/41977521502?pr=1601

lsegal:
(oops earlier comment was incorrect)

We would need to support ruby 2.x+ even if we can't run CI tests against them. CI doesn't represent what versions of ruby yard supports.

seuros:
Remember that dropping support doesn't mean that version 2.x cannot install yard suddenly. 

It just means that future versions won't be installable.

Those legacy versions don't even install in recent hardware like Apple Silicon, and lot it dependencies are EOL or have huge security problem (eg: openssl 3 compatibility).

lsegal:
Marking this as closed. Upon further review this is not a needed change in YARD; it seems to drop support for older RDoc versions for no reason even though they continue to work in YARD.

YARD does not drop support just because libraries are old. rdoc < 6 is still installable on plenty of machines, including in Ruby 4.x:

```
# Gemfile
gem 'rdoc', '< 6'
gem 'e2mmap'
gem 'irb', '= 0.9.6'
```

```
$ bundle && bundle show rdoc && bundle exec ruby --version && bundle exec yard
Bundle complete! 3 Gemfile dependencies, 4 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
/Users/loren/.gem/gems/rdoc-5.1.0
ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin24]
Files:         195
Modules:        52 (    6 undocumented)
Classes:       218 (   41 undocumented)
Constants:     130 (   54 undocumented)
Attributes:    184 (    0 undocumented)
Methods:       986 (  141 undocumented)
 84.59% documented
```

PR #1616, opened 2025-05-11
Only load from `cgi` what is required for Ruby 3.5

Earlopain:
# Description

In Ruby 3.5 most of the cgi gem will be removed. Only the various escape/unescape methods will be retained by default.

But:
* On Ruby 3.5, `cgi/escape` is enough
* Earlier versions need `cgi/util` since the unescape methods don't work otherwise.

I tested this down to Ruby 2.0.

https://bugs.ruby-lang.org/issues/21258

# Completed Tasks

### DIFF
--- a/lib/yard/templates/helpers/html_helper.rb
+++ b/lib/yard/templates/helpers/html_helper.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
-require 'cgi'
+if RUBY_VERSION < '3.5'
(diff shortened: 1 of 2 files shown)


lsegal:
Awesome fix, thanks!

mensfeld:
@lsegal any chance for a release with this? I recently developed https://github.com/mensfeld/yard-lint and stumbled upon this issue with ruby 3.5 :sweat:

PR #1622, opened 2025-08-04
Allow literal symbols in types

apiology:
# Description

This matches behavior of https://yardoc.org/types. Literal symbols are also supported by RBS, sord and recent versions of Solargraph, so this also reduces interoperability friction between tools.  

For instance, valid Solargraph annotations will cause errors in rubocop-yard, which uses YARD for parsing:

```
  lib/solargraph/complex_type/unique_type.rb:200:27: C: YARD/TagTypeSyntax: (SyntaxError) invalid character at :
```

# Completed Tasks

- [X] I have read the [Contributing Guide][contrib].
- [X] The pull request is complete (implemented / written).
- [X] Git commits have been cleaned up (squash WIP / revert commits).
- [X] I wrote tests and ran `bundle exec rake` locally (if code is attached to PR).

[contrib]: https://github.com/lsegal/yard/blob/main/CONTRIBUTING.md

PR #1628, opened 2025-09-13
rubocop-yard: Add support for string literals in TypesExplainer

### LINKED ISSUE #1627: Support symbol and string literal values
Hi, 

I'm coming from an issue that I've opened in `rubocop-yard` here https://github.com/ksss/rubocop-yard/issues/35

I would like to be able to use `@param [:some_symbol, :some_other_symbol]` or `@param ['some_string', 'some_other_symbol']` without having rubocop complain. I mainly rely on YARD for better solargraph LSP code completion/navigation.

The reason why the plugin complains, apparently, is that it depends on `YARD::Tags::TypesExplainer`, which does not seem to support it.

Strangely enough, when I tried it on https://yardoc.org/types.html - which, as mentioned, it uses a [similar implementation](https://github.com/lsegal/yard-types-parser/blob/master/lib/yard_types_parser.rb) to [TypesExplainer](https://github.com/lsegal/yard/blob/5b93b3a4f15a52f3165edef112e718149d294c38/lib/yard/tags/types_explainer.rb#L6) - it seems to work as expected? (although I don't understand how looking at the code)

Image

Would this make sense?

lekemula:
# Description

Resolves #1627

- Add LITERALMATCH regex constant to match symbol (:symbol) and string ('string', "string") literals
- Update TypesExplainer parser to recognize symbol and string literal types
- Format literal values as "a literal value :symbol" for better readability
- Add comprehensive test coverage for LITERALMATCH constant and literal type parsing
- Fixes parsing inconsistency between YARD TypesExplainer and online parser

🤖 Generated with [Claude Code](https://claude.ai/code)

# Completed Tasks

### DIFF
--- a/lib/yard/code_objects/base.rb
+++ b/lib/yard/code_objects/base.rb
@@ -63,6 +63,9 @@ module YARD
     # Regular expression to match a fully qualified method def (self.foo, Class.foo).
     METHODMATCH = /(?:(?:#{NAMESPACEMATCH}|[a-z]\w*)\s*(?:#{CSEPQ}|#{NSEPQ})\s*)?#{METHODNAMEMATCH}/
 
+    # Regular expression to match symbol and string literals
+    LITERALMATCH = /:\w+|'[^']*'|"[^"]*"/
+
     # All builtin Ruby exception classes for inheritance tree.
     BUILTIN_EXCEPTIONS = ["ArgumentError", "ClosedQueueError", "EncodingError",
       "EOFError", "Exception", "FiberError", "FloatDomainError", "IndexError",
--- a/lib/yard/code_objects/base.rb
+++ b/lib/yard/code_objects/base.rb
@@ -63,9 +63,6 @@ module YARD
     # Regular expression to match a fully qualified method def (self.foo, Class.foo).
     METHODMATCH = /(?:(?:#{NAMESPACEMATCH}|[a-z]\w*)\s*(?:#{CSEPQ}|#{NSEPQ})\s*)?#{METHODNAMEMATCH}/
 
-    # Regular expression to match symbol and string literals
-    LITERALMATCH = /:\w+|'[^']*'|"[^"]*"/
-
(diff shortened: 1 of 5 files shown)


lib/yard/code_objects/base.rb
```diff
@@ -63,6 +63,9 @@ def push(value)
     # Regular expression to match a fully qualified method def (self.foo, Class.foo).
     METHODMATCH = /(?:(?:#{NAMESPACEMATCH}|[a-z]\w*)\s*(?:#{CSEPQ}|#{NSEPQ})\s*)?#{METHODNAMEMATCH}/
 
+    # Regular expression to match symbol and string literals
+    LITERALMATCH = /:\w+|'[^']*'|"[^"]*"/
```
lekemula, line 67: Does this constant make sense here?

lsegal (changes requested):
I appreciate the attempt but the LITERALMATCH that this is based on seems perilous if used elsewhere.

We have another competing implementation in #1622 that also doesn't quite get this right but at least is less invasive (it's really just a naming change). 

I think it would make more sense to pull that identifier regex into the types_explainer code directly so that the "best case" logic doesn't leak into other more generalized YARD code paths. My worry is that someone would see LITERALMATCH and use it for _general_ code parsing, which would fail quickly. It's fine in the context of type specifier text, but not for blocks of code.

tldr LITERALMATCH is a case of premature generalization and that regex should be specific to the types explainer code, maybe not even a constant at all.

lsegal:
We went the route of pulling #1622 because there was less blast radius. If you want to roll your changes over that, that would be great, but I think this needs a rebase etc.

Thanks!

lekemula:
Thanks for the notice @lsegal - will do!

lekemula:
Hi @lsegal, I rebased my changes, so this PR is ready for re-review 🙏

lsegal:
Thank you for this!

Issue #1496, opened 2023-05-08
Strange behavior in parsing private methods


unurgunite:
There is a strange behavior while parsing singleton methods withing YARD

## Steps to reproduce
Let's check for class methods in plain Ruby:
```ruby
class T
  def m1
    T.m2
  end

  private

  def self.m2
    print 123
  end

  class << self
    private

    def m3; end
  end
end

T.private_methods.include?(:m2) #=> false
T.private_methods.include?(:m3) #=> true
T.m2 #=> 123=> nil
T.m3 #=> 
(irb):19:in `<main>': private method `m3' called for T:Class (NoMethodError)
	from /home/$USER/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/irb-1.6.4/exe/irb:9:in `<top (required)>'
	from /home/$USER/.rbenv/versions/3.2.2/bin/irb:25:in `load'
	from /home/$USER/.rbenv/versions/3.2.2/bin/irb:25:in `<main>'
```
Look, method `T.m3` is private and `T.m2` -- not. There is additional notes from RuboCop report:
```text
RuboCop: `private` (on line 6) does not make singleton methods private. Use `private_class_method` or `private` inside a `class << self` block instead. [Lint/IneffectiveAccessModifier]
```
## Actual output
Here is output within YARD:
```ruby
require 'yard'

code = <<~CODE
  class T
    def m1
      T.m2
    end

    private

    def self.m2
      print 123
    end

    class << self
      private

      def m3; end
    end
  end
CODE

def detect_private_methods(code)
  YARD.parse_string(code)
  YARD::Registry.all(:class).map do |class_obj|
    class_obj.meths(inherited: false).map do |method_obj|
      if method_obj.scope == :class && method_obj.visibility == :private
        puts "Method #{method_obj.name} is a #{method_obj.visibility} #{method_obj.scope} method!"
      end
    end
  end
end

detect_private_methods(code)
# Method m2 is a private class method!
# Method m3 is a private class method!
# => [[nil, nil, nil]]
```

## Expected Output

In expected output only method `T.m3` should be detected and nothing else:
```ruby
require 'yard'

code = <<~CODE
  class T
    def m1
      T.m2
    end

    private

    def self.m2
      print 123
    end

    class << self
      private

      def m3; end
    end
  end
CODE

def detect_private_methods(code)
  YARD.parse_string(code)
  YARD::Registry.all(:class).map do |class_obj|
[...]

PR #1632, opened 2025-11-06
Respect Ruby semantics for bare visibility with singleton defs

unurgunite:
# Description

YARD incorrectly marked singleton method definitions (`def self.x`) as private when a bare private appeared earlier in the class body. This problem was originally described here (https://github.com/lsegal/yard/issues/1496). Ruby semantics say a bare access modifier affects only the current def-target:

- class body: affects instance methods only
- inside `class << self`: affects class methods only
- `def self.x` in a class body remains public unless:
  - `private_class_method` is used, or
  - `private` appears inside `class << self`

This PR brings YARD in line with Ruby.

Behavior change:
- Bare `private`/`public`/`protected` now only affect methods whose scope matches the current def-target.
- `@!visibility` directive continues to apply regardless of method scope (as before), since it’s a documentation control.
- Explicit statements `private :foo`/`public :foo` are unaffected.
- `private_class_method`/`public_class_method` handlers continue to set exact visibilities.

So, the code from the issue now outputs correctly:
```ruby
require 'yard'

code = <<~CODE
  class T
    def m1
      T.m2
    end

    private

    def self.m2
      print 123
    end

    class << self
      private

      def m3; end
    end
  end
CODE

def detect_private_methods(code)
  YARD.parse_string(code)
  YARD::Registry.all(:class).map do |class_obj|
    class_obj.meths(inherited: false).map do |method_obj|
      if method_obj.scope == :class && method_obj.visibility == :private
        puts "Method #{method_obj.name} is a #{method_obj.visibility} #{method_obj.scope} method!"
      end
    end
  end
end

detect_private_methods(code)
# Method m3 is a private class method!
# => [[nil, nil, nil]]
```

# Completed Tasks

### DIFF
--- a/lib/yard/handlers/base.rb
+++ b/lib/yard/handlers/base.rb
@@ -462,6 +462,18 @@ module YARD
           end
         end
 
+        if docstring.is_a?(String)
+          if (m = docstring.match(/^\s*@!?visibility\s+(public|private|protected)\b/m))
+            vis_sym = m[1].to_sym
+
+            if object.nil?
+              globals.visibility_origin = :directive
+            elsif object.is_a?(CodeObjects::MethodObject)
+              object.visibility = vis_sym
+            end
+          end
+        end
+
         register_transitive_tags(object)
       end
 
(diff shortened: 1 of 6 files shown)


lsegal:
Thanks for this, sorry for the delay!

PR #1646, opened 2026-02-25
fix: relative URL resolution for client-side docs navigation

GunniBusch:
## Summary
Sidebar navigation links are generated as relative URLs that work from the docs root, but they break when the current page is under a nested path.

This change uses the existing `relpath` value https://github.com/lsegal/yard/blob/eddd10c3948021c34a887db403824ce5a892708b/templates/default/layout/html/script_setup.erb#L3 to build a resolved URL (`rel_url = relpath + e.data.url`) and then uses that resolved URL consistently during navigation.

Please see the homebrew ruby docs as an example: https://docs.brew.sh/rubydoc/Formula/FormulaConflict.html. From this class try going to any other class. Also see https://github.com/Homebrew/brew/pull/21589

### DIFF
--- a/templates/default/fulldoc/html/js/app.js
+++ b/templates/default/fulldoc/html/js/app.js
@@ -347,7 +347,8 @@ window.addEventListener(
   "message",
   async (e) => {
     if (e.data.action === "navigate") {
-      const response = await fetch(e.data.url);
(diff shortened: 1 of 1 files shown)


lsegal:
Fixed by ae63056296bd9a47a922361ac19226ca2b12d360

PR #1647, opened 2026-02-25
Fix static HTML class list navigation path duplication (#1639)

### LINKED ISSUE #1639: Class List links generate incorrect relative paths in yard 0.9.38 (static HTML)
### Summary

In yard 0.9.38, links in the Class List of generated static HTML sometimes point to incorrect paths and result in 404 errors.

This issue does not occur in yard 0.9.37 and seems to be a regression introduced in yard 0.9.38.

### Environment

* Ruby: (any)
* yard: 0.9.38
* Web server: WEBrick (static file serving)

`Gemfile`
```ruby
source "https://rubygems.org"

gem "webrick"
gem "yard", "0.9.38"
```

`.yardopts`
```
src/**/*.rb
```

### Reproduction steps

`src/main.rb`
```ruby
class Foo
  class Bar; end
  class Baz; end
end
```

Generate documentation:
```console
bundle exec yard doc
```

Serve generated files as static HTML:
```console
cd doc/
ruby -rwebrick -e 'WEBrick::HTTPServer.new(DocumentRoot: ".", Port: 3000).start'
```

1. Open `http://localhost:3000/Foo/Bar.html`
2. In the Class List, click `Baz < Object`

### Expected behavior

The link should navigate to:
```
http://localhost:3000/Foo/Baz.html
```

### Actual behavior

The browser tries to access:
```
http://localhost:3000/Foo/Foo/Baz.html
```

which results in a 404 error.
It appears that the relative path in the Class List is resolved incorrectly when viewing a nested class page.

### Screenshots

Class List link being clicked:

Image

Resulting network request (404):

Image

### Additional notes / workarounds

The issue can be avoided by one of the following approaches:

* Use yard 0.9.37
* Use yard based on 0.9.38 with commit e18b8440d3b580df3c295addfde605addac754b6 reverted
* Use `yard server` instead of serving generated HTML manually
  * (I understand this is not suitable when exporting static HTML, but mentioning it for reference)

### Notes

This issue only appears when serving the generated documentation as **static HTML**.
It does not occur when using `yard server`.

carsonclarkemagrab-toast:
## Summary
Fixes #1639

This fixes a breaking bug in v0.9.38 where clicking nested class links in the class list generates incorrect URLs when serving static HTML documentation, resulting in 404 errors.

## Root Cause
The fetch-based navigation system (commit e18b8440) resolves relative URLs from the class list iframe against the main frame's current URL instead of the iframe's base URL. When viewing `/Foo/Bar.html` and clicking a link to `Foo/Baz.html`, the browser incorrectly resolves it as `/Foo/Foo/Baz.html`.

## The Fix
* Resolve URLs against the iframe's `contentWindow.location` instead of the main frame location
* This ensures relative paths like `Foo/Baz.html` resolve correctly regardless of the current page

## Testing

Reproduced the issue with nested class structure:
```ruby
class Foo
    class Bar; end
    class Baz; end
end
```

- Before fix: Clicking Foo::Baz from /Foo/Bar.html → /Foo/Foo/Baz.html
- After fix: Clicking Foo::Baz from /Foo/Bar.html → /Foo/Baz.html

Related: https://github.com/block/elasticgraph/issues/1048

### DIFF
--- a/templates/default/fulldoc/html/js/app.js
+++ b/templates/default/fulldoc/html/js/app.js
@@ -347,7 +347,12 @@ window.addEventListener(
   "message",
   async (e) => {
     if (e.data.action === "navigate") {
-      const response = await fetch(e.data.url);
+      // Resolve the URL against the iframe's location to handle relative paths correctly
+      const iframe = document.getElementById("nav");
+      const iframeUrl = iframe ? iframe.contentWindow.location.href : window.location.href;
+      const resolvedUrl = new URL(e.data.url, iframeUrl).href;
+
+      const response = await fetch(resolvedUrl);
       const text = await response.text();
       const parser = new DOMParser();
       const doc = parser.parseFromString(text, "text/html");
@@ -383,12 +388,12 @@ window.addEventListener(
 
       document.getElementById("class_list_link").classList = classListLink;
 
-      const url = new URL(e.data.url, "https://localhost");
+      const url = new URL(resolvedUrl);
       const hash = decodeURIComponent(url.hash ?? "");
       if (hash) {
         document.getElementById(hash.substring(1)).scrollIntoView();
       }
-      history.pushState({}, document.title, e.data.url);
+      history.pushState({}, document.title, resolvedUrl);
     }
   },
   false


lsegal:
This was fixed in v0.9.40

PR #1662, opened 2026-03-21
Allow using Prism for parsing

kddnewton:
Hi there!

I wanted to open this PR to discuss if you're interested in using the Prism parser instead of Ripper. This is a WIP branch that introduces the ability to do that via an ENV var. It also includes some benchmarks. Locally I'm seeing about a 300% performance improvement through this work.

It's not done, and some edge case tests are still failing, and I'm sure there's lots to work through still. But I wanted to pause here because this is becoming quite a lot of code to review. So I'd love to know if you're interested in this direction before I continue down the path.

I also have been purposefully avoiding sharing code between the processors to make this diff smaller (which is in vain at this point...), but if this were to truly go through, I think it would make more sense to start factoring stuff out into shared classes.

Let me know what you think.

### DIFF
--- a/benchmark/prism_vs_ripper.rb
+++ b/benchmark/prism_vs_ripper.rb
@@ -1,0 +1,162 @@
+# frozen_string_literal: true
+#
+# Benchmark comparing Prism and Ripper parser backends for YARD.
+#
+# Usage:
+#   ruby benchmark/prism_vs_ripper.rb
+#   ruby benchmark/prism_vs_ripper.rb path/to/file.rb   # single file
+#   ruby benchmark/prism_vs_ripper.rb path/to/dir        # directory
+
+require "benchmark"
+require "stringio"
+$LOAD_PATH.unshift File.expand_path("../lib", __dir__)
+begin require "prism"; rescue LoadError; end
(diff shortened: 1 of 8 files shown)


lsegal:
Curious if you have plans to take these changes up. Adding a new parser library is not a simple slot-in, YARD has an extensibility and plugin layer that requires quite a bit of care to work with. If the plan is to completely migrate in a backward incompatible way, it would need to be well documented with migration paths for consumers. To make this a bit more complex, we _already_ have a "legacy" parser stack, so that would need to be renamed, ripper moved, etc.

kddnewton:
Hey @lsegal — yes it's on my list, but honestly it's going to be a while. This was definitely the largest project I have attempted to migrate, and clearly I was missing some context :). I will do my best to come up with a design that works for everyone, be it completely backward compatible or with a very clear migration process.

omarqureshi:
@kddnewton - happy to work with you on this. I'm currently writing a jsii target in Ruby - the documentation spans across nearly 12,000 files so YARD runs take around an hour

kddnewton:
Hi @omarqureshi — you're welcome to pick this up if you like but I won't have time to return to it for a while

PR #1671, opened 2026-04-09
Fix bracket/brace map corruption from Ruby 3.0+ pattern matching deconstruction

### LINKED ISSUE #1547: Bug: deconstruction + array of symbols
# Description

YARD is losing control with deconstruction and array of symbols.

## Steps to reproduce

This is the minimal reproduction for the issue.

`Gemfile`:
```ruby
source 'https://rubygems.org'
ruby '>= 3.0'
gem 'yard'
```

`.yardopts`:
```
yardoc
--output-dir rdoc
--markup=markdown
src/**/*.rb
```

`src/main.rb`:
```ruby
module Mod
  class A
    # Some fn
    # @param arg [Object]
    def fn(arg)
      case arg
      in SomeClass['array', v]
        '42'
      else
        super
      end
    end

    # A nice little const.
    CONST = %i[a b c]
  end
end
```

## Actual Output

image

## Expected Output

image

And that was something I can achieve by:

1. Putting the const before the fn
  ```ruby
  module Mod
    class A
      # A nice little const.
      CONST = %i[a b c]

      # Some fn
      # @param arg [Object]
      def fn(arg)
        case arg
        in SomeClass['array', v]
          '42'
        else
          super
        end
      end
    end
  end
  ```
2. Avoiding `%i` notation:
  ```ruby
  module Mod
    class A
      # Some fn
      # @param arg [Object]
      def fn(arg)
        case arg
        in SomeClass['array', v]
          '42'
        else
          super
        end
      end

      # A nice little const.
      CONST = [:a, :b, :c]
    end
  end
  ```

None of those solutions are actually acceptable in my case:
1. I cannot change the order of these particular things, because the deconstruction happens in a class different than that of the const, and I cannot extract the class to a separate file (I can —physically— of course, it's just that it doesn't make sense in my project).
1. I cannot avoid `%i` notation because the actual const is humongous (again, I physically can, but I would be violating the codebase coding style to please YARD).

## Environment details:

- OS: mac (M1) or Linux (ubuntu) … I don't think it matters
- Ruby version (`ruby -v`): 3.3.1
- YARD version (`yard -v`): 0.9.36
- Relevant software dependency/versions:
  - none

dduugg:
# Description

In Ruby 3.0+ pattern matching, bracket-based and braced patterns fire scanner events for their delimiters but do not fire the same parser events as ordinary array indexing or hash literals. YARD had no handlers for these pattern-match parser events, so entries were pushed to the bracket/brace maps but never consumed. Since those maps are read FIFO/LIFO by later expressions, the stale entries produced garbled source ranges on the next array indexing or hash literal after a pattern match.

Three event handlers are added:

- **`on_aryptn`**: Array patterns (`Bar['key', v]`, `Bar[a, b]`, bare `[a, b]`) fire `on_aryptn` instead of `on_aref`. The handler pops from `@map[:lbracket]` and shifts from `@map[:aref]`, mirroring what `on_aref` would have done.
- **`on_fndptn`**: Find patterns (`Bar[*, v, *]`) fire `on_fndptn` instead of `on_aref`. Same bracket map cleanup.
- **`on_hshptn`**: Braced hash patterns (`{ key: val }`) fire `on_hshptn` instead of `on_hash`, leaving a stale entry in `@map[:lbrace]`. To distinguish braced patterns from bare hash patterns (`key: val`, which fire no brace scanner events), `on_rbrace` is overridden to track closing brace positions in `@map[:rbrace]`. `on_hshptn` cleans up both maps only when `@map[:rbrace]` confirms a closing brace was scanned, using `||=` guards on both map accesses for consistency with the rest of the map handling code. `visit_event` is updated to drain `@map[:rbrace]` when consuming a `@map[:lbrace]` entry, keeping the maps in sync for normal hash literals and brace blocks (including the case where a braced pattern appears inside a brace block).

Fixes #1547

# Completed Tasks

-----------------

# Description

# Completed Tasks

### DIFF
--- a/lib/yard/parser/ruby/ruby_parser.rb
+++ b/lib/yard/parser/ruby/ruby_parser.rb
@@ -244,6 +244,7 @@ module YARD
             sstart = child_node.source_range.first
           else
             lstart, sstart = *(map ? map.pop : [lineno, @ns_charno - 1])
+            (@map[:rbrace] ||= []).shift if map && MAPPINGS[node.type] == :lbrace
           end
 
           raise "Cannot determine start of node #{node} around #{file}:#{lineno}" if lstart.nil? || sstart.nil?
@@ -336,6 +337,7 @@ module YARD
         undef on_aref_field
         undef on_lbracket
         undef on_rbracket
+        undef on_rbrace
         undef on_string_literal
         undef on_lambda
         undef on_unary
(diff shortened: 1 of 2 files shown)


lsegal:
Thanks for the contribution!

Issue #1116, opened 2017-09-01
Raise StackLevelTooDeep when defined cycle includes modules


HaiTo:
run `yard doc *.rb`  and raise StackLevelTooDeep.
As a result I can not get the text.

## Steps to reproduce

 1. define files ↓

```rb
# in one.rb
module One
  include Two
end

module Two
  include One
end
```

 2. run `yard doc one.rb`

 3. raise `StackLevelTooDeep`

## Actual Output
```
$ yard doc one.rb --debug
[debug]: Parsing ["one.rb"] with `ruby` parser
[debug]: Parsing one.rb
[debug]: Serializing to .yardoc/objects/root.dat
[debug]: Re-generating object ...
[debug]: Re-generating object One...
[debug]: Re-generating object Two...
[debug]: Generating asset js/jquery.js
[debug]: Serializing to doc/js/jquery.js
[debug]: Generating asset js/app.js
[debug]: Serializing to doc/js/app.js
[debug]: Generating asset js/full_list.js
[debug]: Serializing to doc/js/full_list.js
[debug]: Generating asset css/style.css
[debug]: Serializing to doc/css/style.css
[debug]: Generating asset css/common.css
[debug]: Serializing to doc/css/common.css
[debug]: Generating asset css/full_list.css
[debug]: Serializing to doc/css/full_list.css
[debug]: Generating asset class_list.html
[debug]: Serializing to doc/class_list.html
[debug]: Generating asset method_list.html
[debug]: Serializing to doc/method_list.html
[debug]: Generating asset file_list.html
[debug]: Serializing to doc/file_list.html
[debug]: Generating asset frames.html
[debug]: Serializing to doc/frames.html
[debug]: Serializing to doc/index.html
[debug]: Serializing to doc/_index.html
[debug]: Serializing to doc/top-level-namespace.html
/Users/otsuka.hayato/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/yard-0.9.9/lib/yard/code_objects/proxy.rb:132:in `hash': stack level too deep (SystemStackError)
	from /Users/otsuka.hayato/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/yard-0.9.9/lib/yard/code_objects/namespace_object.rb:197:in `|'
[...]

PR #1672, opened 2026-04-09
Fix false self-referential mixin when bare name matches ancestor namespace

dduugg:
# Description

When a module like `A::Enumerable` contains `include Enumerable`, YARD's lexical resolver walks up the namespace tree, reaches parent namespace `A`, and finds `A::Enumerable` as the match for the bare name `Enumerable`. This records the module as including itself — producing incorrect documentation and potential infinite recursion in `inheritance_tree`.

This is the valid-Ruby case reported in #1116 (comment): https://github.com/lsegal/yard/issues/1116#issuecomment-353475432

```ruby
module A
end

module A::Enumerable
  include Enumerable  # YARD incorrectly resolved this to A::Enumerable itself
end
```

**Fix:** In `RegistryResolver#lookup_by_path`, after a direct-path lookup returns a result while searching in a *parent* namespace, skip that result if it equals the original starting namespace. This breaks the false cycle while leaving all other lookups unaffected (e.g., root-level lookups, qualified references).

# Completed Tasks

### DIFF
--- a/lib/yard/registry_resolver.rb
+++ b/lib/yard/registry_resolver.rb
@@ -75,6 +75,13 @@ module YARD
       lexical_lookup = 0
       while namespace && !resolved
         resolved = lookup_path_direct(namespace, path, type)
+        # Prevent a bare name from resolving back to the namespace we started
+        # from when searching through a parent namespace. For example,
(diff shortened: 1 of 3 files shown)


lsegal:
Great fix, thanks!

PR #1630, opened 2025-09-28
rubocop-yard: Add support for multiple & nested Hash keys definition

lekemula:
# Description

Follow-up on https://github.com/lsegal/yard/pull/1628 (omit first 2 commits when reviewing)

Adds support for multiple and nested hash keys - example from specs:

``` ruby
"Hash{:key_one, :key_two => String; :key_three => Symbol; :key_four => Hash{:sub_key_one => String}}" => "a Hash with keys made of (a literal value :key_one or a literal value :key_two) and values of (Strings) and keys made of (a literal value :key_three) and values of (Symbols) and keys made of (a literal value :key_four) and values of (a Hash with keys made of (a literal value :sub_key_one) and values of (Strings))",
"Hash{:key_one => String, Number; :key_two => String}" => "a Hash with keys made of (a literal value :key_one) and values of (Strings or Numbers) and keys made of (a literal value :key_two) and values of (Strings)"
```

NOTE: These changes probably only impact on https://github.com/ksss/rubocop-yard and have no effect elsewhere.

PS. I think this alongside the base PR, will open up a lot of opportunities for better `Hash` completion/typechecking for [solargraph](https://github.com/castwide/solargraph) gem.

# Completed Tasks

- [x] I have read the [Contributing Guide][contrib].
- [x] The pull request is complete (implemented / written).
- [x] Git commits have been cleaned up (squash WIP / revert commits).
- [x] I wrote tests and ran `bundle exec rake` locally (if code is attached to PR).

[contrib]: https://github.com/lsegal/yard/blob/main/CONTRIBUTING.md

Issue #1689, opened 2026-06-29
rdoc 8.0 was released


akimd:
## Summary

I don't know the details about yard, but it seems that it relies on rdoc.  And since rdoc 8.0, I have:

```
/Users/akim/project/vendor/ruby/3.4.0/gems/rdoc-8.0.0/lib/rdoc/markup/to_html.rb:127:in 'initialize': wrong number of arguments (given 1, expected 0) (ArgumentError)
	from /Users/akim/project/vendor/ruby/3.4.0/bundler/gems/yard-c8a71ef81029/lib/yard/templates/helpers/markup/rdoc_markup.rb:19:in 'YARD::Templates::Helpers::Markup::RDocMarkupToHtml#initialize'
	from /Users/akim/project/vendor/ruby/3.4.0/bundler/gems/yard-c8a71ef81029/lib/yard/templates/helpers/markup/rdoc_markup.rb:45:in 'Class#new'
	from /Users/akim/project/vendor/ruby/3.4.0/bundler/gems/yard-c8a71ef81029/lib/yard/templates/helpers/markup/rdoc_markup.rb:45:in 'block in YARD::Templates::Helpers::Markup::RDocMarkup#initialize'
	from /Users/akim/project/vendor/ruby/3.4.0/bundler/gems/yard-c8a71ef81029/lib/yard/templates/helpers/markup/rdoc_markup.rb:44:in 'Thread::Mutex#synchronize'
	from /Users/akim/project/vendor/ruby/3.4.0/bundler/gems/yard-c8a71ef81029/lib/yard/templates/helpers/markup/rdoc_markup.rb:44:in 'YARD::Templates::Helpers::Markup::RDocMarkup#initialize'
	from /Users/akim/project/vendor/ruby/3.4.0/bundler/gems/yard-c8a71ef81029/lib/yard/templates/helpers/markup/rdoc_markdown.rb:15:in 'YARD::Templates::Helpers::Markup::RDocMarkdown#initialize'
	from /Users/akim/project/vendor/ruby/3.4.0/bundler/gems/yard-c8a71ef81029/lib/yard/templates/helpers/html_helper.rb:95:in 'Class#new'
	from /Users/akim/project/vendor/ruby/3.4.0/bundler/gems/yard-c8a71ef81029/lib/yard/templates/helpers/html_helper.rb:95:in 'YARD::Templates::Helpers::HtmlHelper#html_markup_markdown'
```

If the callstack is not enough, I can try to find a small test case.

[...]

PR #1690, opened 2026-07-11
Fix TypesExplainer folding trailing types into a preceding Hash{...}

### LINKED ISSUE #1688: TypesExplainer::Parser silently folds trailing types into preceding Hash{...} block
## Summary

`YARD::Tags::TypesExplainer::Parser.parse` incorrectly absorbs types that follow a closed `Hash{...}` collection into that hash's value list, instead of treating them as sibling top-level types. This is inconsistent with YARD's own docstring parser, which correctly splits the same text into separate types.

## Reproduction (yard 0.9.44, ruby 3.3.11)

```ruby
require "yard"

# Docstring parser — correctly splits into 2 types:
YARD::DocstringParser.new
  .parse("@return [Hash{Symbol => String}, nil]")
  .tags.first.types
# => ["Hash{Symbol => String}", "nil"]

# TypesExplainer parser on the same joined text — returns 1 explainer:
YARD::Tags::TypesExplainer::Parser.parse("Hash{Symbol => String}, nil")
# => [#<HashCollectionType name="Hash"
#       key_value_pairs=[[[Type "Symbol"], [Type "String", Type "nil"]]]>]
```

The trailing `, nil` gets pulled inside the Hash's value list, even though the `}` already closed it. The same shape with `Array<...>, nil` is parsed correctly — only the `Hash{...}` form mis-parses.

## Expected

Two top-level explainers: one `HashCollectionType` and one `Type "nil"`, matching what the docstring parser produces.

## Impact

Downstream tools that rely on `TypesExplainer` get a corrupted type tree. In particular, `rubocop-yard`'s `YARD/CollectionStyle` cop crashes and silently rewrites user code based on the wrong tree (filed separately at ksss/rubocop-yard).

lekemula:
# Description

Fixes #1688.

`YARD::Tags::TypesExplainer::Parser.parse` absorbed types that follow a closed `Hash{...}` collection into that hash's value list, instead of treating them as sibling top-level types — diverging from `DocstringParser`, which splits the same text correctly:

```ruby
# Before
YARD::Tags::TypesExplainer::Parser.parse("Hash{Symbol => String}, nil").size # => 1 (wrong)
# After
YARD::Tags::TypesExplainer::Parser.parse("Hash{Symbol => String}, nil").size # => 2
YARD::DocstringParser.new.parse("@return [Hash{Symbol => String}, nil]").tags.first.types
# => ["Hash{Symbol => String}", "nil"]
```

This was a regression from #1630 (which moved hash parsing into its own nested loop).

## Root cause

In `parse_hash_collection`, the value sub-parse did **not** stop on the hash's closing `}` (`:hash_collection_end`). Since the main `parse` case statement also doesn't handle that token, the `}` was silently consumed and the value parse kept going, swallowing the trailing `, nil`. It only surfaced when a type followed the `}` — `Array<...>, nil` works because collection parsing is inlined in `parse` and returns control to the same loop, whereas the hash runs in its own nested loop.

## Fix

- Stop the value sub-parse on `:hash_collection_end` (in addition to `;` and end-of-input).
- Track the terminating token in `parse`; finish the hash when the value parse ended on `}` or end-of-input, returning control to the parent `parse` loop (which then reads the sibling `, nil`). A `;` terminator still continues to the next key group.

Multi-pair (`;`), nested-hash, and hash-as-last-type cases are unchanged (covered by existing specs). Added regression specs for the sibling-after-hash case.

# Completed Tasks

- [x] I wrote tests and ran the suite locally (`bundle exec rspec spec/tags/ spec/docstring_parser_spec.rb` — 181 examples, 0 failures).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/yard/tags/types_explainer.rb
+++ b/lib/yard/tags/types_explainer.rb
@@ -186,6 +186,7 @@ module YARD
               raise SyntaxError, "expecting name, got '#{token}'" if name.nil?
               type = create_type(name) unless type
               current_parsed_types << type
+              @terminator_token = token_type
               finished = true
             when :type_name
               raise SyntaxError, "expecting END, got name '#{token}'" if name
@@ -245,9 +246,12 @@ module YARD
             when :hash_collection_value
               # => - current keys map to the next value(s)
               raise SyntaxError, "no keys before =>" if current_keys.empty?
-              values = parse(until_tokens: [:hash_collection_value_end, :parse_end])
+              values = parse(until_tokens: [:hash_collection_value_end, :hash_collection_end, :parse_end])
               key_value_pairs << [current_keys, values]
               current_keys = []
+              # The value parse consumes its terminator_token. A closing '}' (or end of input)
+              # ends the hash; ';' continues to the next key group.
+              finished = true if @terminator_token == :hash_collection_end || @terminator_token == :parse_end
             when :hash_collection_end, :parse_end
               # End of hash
               finished = true
--- a/spec/tags/types_explainer_spec.rb
+++ b/spec/tags/types_explainer_spec.rb
@@ -213,6 +213,16 @@ RSpec.describe YARD::Tags::TypesExplainer do
       expect(type.map(&:name)).to eq ['false', 'true', 'nil', '4', ':foo']
     end
 
+    it "parses a type following a closed hash collection as a sibling" do
+      type = parse("Hash{Symbol => String}, nil")
+      expect(type.size).to eq 2
+      expect(type[0]).to be_a(YARD::Tags::TypesExplainer::HashCollectionType)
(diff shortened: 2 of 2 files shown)


lekemula:
@lsegal Some Ruby version failures seem to be related to https://github.com/lsegal/yard/issues/1689

The others are green ✅ 

image

lsegal:
Sorry I missed this fix when having my workflow loops look at this issue, apologies but I believe #1688 is correctly fixed now. Thank you for the contribution though!

lekemula:
> Sorry I missed this fix when having my workflow loops look at this issue, apologies but I believe #1688 is correctly fixed now. Thank you for the contribution though!

No worries at all - happy to see it fixed and sorry for the oversight!

PR #1695, opened 2026-07-14
Use @param types for Struct and Data members

### LINKED ISSUE #1684: Recognize @param on Struct.new / Data.define constant assignments
## Summary

First, an honest admission: this discussion probably should have started here in YARD before we made changes downstream in Solargraph — but we were eager to ship some kind of support sooner rather than later, and here we are. This issue is to ask retroactively whether it makes sense for YARD itself to natively recognize `@param` annotations on `Struct.new` / `Data.define` constant assignments as documenting the synthesized accessors.

## Problem statement

To document the fields of a `Struct.new` / `Data.define` today in a way YARD actually picks up, you have to use `@!attribute` per field. That's a lot of repetition for what is conceptually a flat list of name → type:

```ruby
class Consent < Data.define(:consent_type_id, :consent_type_name, :value) # rubocop:disable Style/DataInheritance
  # @!attribute [r] consent_type_id
  #   @return [String]
  # @!attribute [r] consent_type_name
  #   @return [String]
  # @!attribute [r] value
  #   @return [Boolean]
end
```

Two pain points here:

1. **Verbose, repetitive boilerplate.** Two lines of YARD per field (`@!attribute` + `@return`), even though the name is already declared in the `Data.define(...)` call right next to it.
2. **Forces the class-inheritance form.** To attach `@!attribute` blocks at all, you have to open a class body — which means writing `class Consent < Data.define(...)`. That form is itself flagged by RuboCop's `Style/DataInheritance` and requires `# rubocop:disable Style/DataInheritance` just to document the type. The idiomatic `Consent = Data.define(...)` form has nowhere to hang the docs.

What developers naturally reach for is the concise form:

```ruby
# @param consent_type_id [String]
# @param consent_type_name [String]
# @param value [Boolean]
Consent = Data.define(:consent_type_id, :consent_type_name, :value)
```

…but YARD doesn't currently associate those `@param` tags with the generated accessors, so they're effectively no-ops in YARD's object graph. Rationale here for choosing `@param` was that, in esence `Data/Struct` definitions are mostly a syntactic sugar over`#initialize` and `#attr_reader` or `#attr_accessor`. 

While Solargraph has added its own "opinionated" handling for this pattern ([castwide/solargraph#970](https://github.com/castwide/solargraph/pull/970), [castwide/solargraph#939](https://github.com/castwide/solargraph/pull/939)), **gems can't really make use of it** — gem documentation, type information, and lint tooling all flow primarily through YARD's parser, not through Solargraph. So a Solargraph-only fix helps editor users but leaves the gem ecosystem (documentation sites, RBS generation, lint tools) still treating these annotations as no-ops. Lint rules in particular even flag them as "meaningless," because YARD's object graph doesn't know what they mean:

- [ksss/rubocop-yard#36](https://github.com/ksss/rubocop-yard/issues/36) (fix in flight: [ksss/rubocop-yard#42](https://github.com/ksss/rubocop-yard/pull/42))
- [mensfeld/yard-lint#152](https://github.com/mensfeld/yard-lint/issues/152)

Each tool ends up special-casing `Struct.new` / `Data.define` on its own — N tools writing N versions of the same heuristic, each drifting in their own way.

## Proposed solution

When YARD parses a constant assignment whose RHS is `Struct.new(...)` or `Data.define(...)` (with or without a block), recognize any `@param name [Type]` tags on the preceding comment block and propagate them as the documented type of the corresponding generated accessor (and ideally the initializer parameter). Concretely:

- For `Consent = Data.define(:consent_type_id, :consent_type_name, :value)` with `@param consent_type_id [String]`, the synthesized `Consent#consent_type_id` reader would be documented with `@return [String]`.
- The same for `Struct.new`, including the writer for non-readonly fields and the corresponding `initialize` parameter.
- Constants whose RHS is *not* `Struct.new` / `Data.define` are unchanged.

The result is the equivalent of all that `@!attribute` boilerplate, but written once per field and without forcing the developer into a form RuboCop fights.

## Alternatives considered

- **Keep the status quo (`@!attribute` per field).** Works, but verbose, and pushes developers toward the `class Consent < Data.define(...)` form that triggers `Style/DataInheritance`.
- **Each downstream tool handles it on its own.** This is where we are today. Works, but ends up duplicated across Solargraph, rubocop-yard, yard-lint, and any future YARD-consuming tool, each with subtly different rules — and as noted above, Solargraph's handling doesn't reach the wider gem/doc tooling ecosystem.
- **A plugin or extension.** Possible, but if YARD treated this natively it would benefit every downstream consumer for free.

## Scope and impact

- **Who it helps:** anyone documenting `Struct.new` / `Data.define` constants (very common in modern Ruby, especially since `Data` shipped in 3.2), plus every YARD-consuming tool (Solargraph, RBS generation, lint tools, IDE integrations).
- **Behavior change:** additive — `@param` tags currently silently ignored on these constants would start being associated with the generated members. Existing documented Structs without `@param` tags are unaffected. Existing `@!attribute`-style docs keep working.
- **Willing to contribute:** yes, happy to draft a PR if the maintainers think this is in scope for YARD itself.

## Additional context

- Solargraph PRs that motivated this: [castwide/solargraph#970](https://github.com/castwide/solargraph/pull/970), [castwide/solargraph#939](https://github.com/castwide/solargraph/pull/939)
- Lint-side issues caused by YARD not modeling this: [ksss/rubocop-yard#36](https://github.com/ksss/rubocop-yard/issues/36), [mensfeld/yard-lint#152](https://github.com/mensfeld/yard-lint/issues/152)
- Related lint-side fix: [ksss/rubocop-yard#42](https://github.com/ksss/rubocop-yard/pull/42)
- RuboCop rule the current workaround collides with: `Style/DataInheritance`
- Older discussions / workarounds in the ecosystem:
  - [lsegal/yard#766](https://github.com/lsegal/yard/issues/766) — earlier request for Struct documentation support in YARD itself
  - [rspec/rspec-core#1765](https://github.com/rspec/rspec-core/issues/1765) — the "reopen the constant as a class" workaround (`MyClass = Struct.new(...); class MyClass; def ...; end`) used in the wild to get YARD to pick anything up

I would be more than happy to draft something, if you find this proposal reasonable and acceptable.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

lsegal:
# Description

`Struct.new` and `Data.define` constant assignments already create generated member methods, but their handler only consulted `@attr*` annotations. As a result, `@param` types stayed on the generated class, the accessors defaulted to `Object`, and reprocessing the class comment for each generated method emitted false unknown-parameter warnings.

This change uses a matching `@param` tag as the generated member's type source while preserving existing `@attr*` precedence. Struct readers, writers, and writer return values now carry the declared type; Data readers do the same. Generated member registration no longer reapplies the class docstring, so the concise annotation form is warning-free while keeping the standard generated descriptions.

The unreleased changelog documents the new behavior.

# User impact

Developers can document idiomatic `Struct.new` and `Data.define` constant assignments with one `@param name [Type]` line per member. YARD consumers now see those types on the synthesized accessors without requiring `@!attribute` boilerplate.

# Tests

- `bundle exec rspec spec/handlers/constant_handler_spec.rb spec/handlers/class_handler_spec.rb`
- `bundle exec rake` (2,788 examples, 0 failures; documentation warning check passed)

# Completed Tasks

- [x] I have read the [Contributing Guide](https://github.com/lsegal/yard/blob/main/CONTRIBUTING.md).
- [x] Git commits have been cleaned up.
- [x] I wrote tests and ran `bundle exec rake` locally.

Closes #1684

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # main
 
+- Use `@param` types to document generated `Struct.new` and `Data.define` members (#1684)
 - Add compatibility with RDoc 8
 - Fix TypesExplainer parsing of types following Hash collections (#1688)
 - Fix HTML generation for RBS constants without source values (#1686)
--- a/lib/yard/handlers/ruby/struct_handler_methods.rb
+++ b/lib/yard/handlers/ruby/struct_handler_methods.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-# Helper methods to parse @attr_* tags on a class.
+# Helper methods to document generated Struct and Data members.
 #
 # @deprecated The use of +@attr+ tags are deprecated since 0.8.0 in favour of
 #   the +@!attribute+ directive. This module should not be relied on.
@@ -17,6 +17,26 @@ module YARD::Handlers::Ruby::StructHandlerMethods
   def member_tag_for_member(klass, member, type = :read)
     specific_tag = type == :read ? :attr_reader : :attr_writer
     (klass.tags(specific_tag) + klass.tags(:attr)).find {|tag| tag.name == member }
+  end
+
+  # Extracts the user's defined @param tag for a given generated member.
+  #
+  # @param [ClassObject] klass the class whose tags we're searching
+  # @param [String] member the name of the struct or data member we need
+  # @return [Tags::Tag, nil] the matching tag, or nil if not found
+  def parameter_tag_for_member(klass, member)
+    klass.tags(:param).find {|tag| tag.name == member }
+  end
+
+  # Returns the tag that supplies a generated member's type. Existing @attr*
+  # tags take precedence over the more concise @param form.
+  #
+  # @param [ClassObject] klass the class whose tags we're searching
+  # @param [String] member the name of the struct or data member we need
+  # @param [Symbol] type reader method, or writer method?
+  # @return [Tags::Tag, nil] the tag supplying the type, or nil if not found
+  def type_tag_for_member(klass, member, type = :read)
+    member_tag_for_member(klass, member, type) || parameter_tag_for_member(klass, member)
   end
 
   # Retrieves all members defined in @attr* tags
@@ -61,7 +81,7 @@ module YARD::Handlers::Ruby::StructHandlerMethods
   # @return [String] a docstring to be attached to the getter method for this member
   def add_reader_tags(klass, new_method, member)
     member_tag = member_tag_for_member(klass, member, :read)
-    return_type = return_type_from_tag(member_tag)
+    return_type = return_type_from_tag(type_tag_for_member(klass, member, :read))
     getter_doc_text = member_tag ? member_tag.text : "Returns the value of attribute #{member}"
     new_method.docstring.replace(getter_doc_text)
     new_method.add_tag YARD::Tags::Tag.new(:return, "the current value of #{member}", return_type)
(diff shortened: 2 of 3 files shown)


lekemula:
@lsegal appreciate you accepting the proposal! 

I was wondering though, whether it would make sense to add some documentation in `docs/Tags.md` maybe somewhere near https://rubydoc.info/gems/yard/file/docs/Tags.md#param, as this is quite "a special usage" of `@param`?

lsegal:
I think it would make sense to have param docs for this, yes.

PR #1700, opened 2026-07-31
Add intersection ("A & B") and union ("A | B" / "[A | B]") type syntax

### LINKED ISSUE #1644: Syntax for intersection types
I'm looking to support intersection types in YARD tags in Solargraph.

Do you have preferences on syntax?  My first inclination is `Foo & Bar`, but we could also use something like `intersection<Foo, Bar>` or `both<Foo, Bar>` - one of the latter two might be convenient if you'd like to avoid impact on YARD's implementation.

**Background:**

Currently [Foo, Bar, #to_baz] in YARD is a union type - when used as type on a method parameter, that means the parameter is/imports/inherits Foo, is/imports/inherits Bar, or quacks the foo method.

Similarly, `Foo | Bar | _ToBaz` in RBS is a a union type.   RBS gives duck types a name for richer definitions - let's pretend `_ToBaz` is one of those.  See 'interfaces' in https://github.com/ruby/rbs/blob/master/docs/syntax.md

RBS supports intersection type with the '&' operator - so `Foo & Bar & _ToBaz` in RBS, when used as a method parameter, means that the argument is/imports/inherits Foo, is/imports/inherits Bar, AND quacks the _ToBaz type.

### LINKED ISSUE #1699: No general grouping syntax for unions nested inside tuples, intersections, etc.
*🤖 Filed by Claude, not the account owner — acting on their behalf via their GitHub credentials.*

Split out from a comment on #1644, where this came up while implementing intersection types in Solargraph (https://github.com/castwide/solargraph/pull/1231) — but on closer inspection it's not intersection-specific. It's a general gap: **there's no syntax to group a union so it can be nested inside a construct that already gives `,` a different meaning at that position.**

### Minimal example (no intersections involved)

Given YARD's existing fixed-tuple-parameter syntax, `Array(Integer, String)` is a 2-element tuple `[Integer, String]` — each comma-separated item is a positional slot.

Now say you want a 1-element tuple whose single element is *either* `Integer` or `String`. There's no way to write that. `,` inside `()` already means "next positional slot," not "alternative for this slot," so you can't reuse it for a union there. And parens can't help either — `Array((Integer, String))` doesn't group `(Integer, String)` as a union in slot 0; it reads as a *nested tuple*, since `()` is unconditionally fixed-tuple syntax. Verified against the current Solargraph implementation:

```
Array(Integer, String)     -> subtypes: [Integer, String]   (2 slots, as expected)
Array((Integer, String))   -> subtypes: [(Integer, String)] (1 slot, holding a NESTED TUPLE, not a union)
```

There is no third notation available to say "these are grouped as alternatives, not sequential."

### Where else this shows up

- **Intersections** (#1644, #1699's origin): RBS allows a union as one member of an intersection (`(A | B) & C` is valid RBS) — same problem, different container. See the discussion on #1644 for detail.
- **Hash/Array/Set generic params** happen to dodge this by convention rather than solving it: `Array<Integer, String>` already means "array of `(Integer | String)`" — multiple params in that (non-fixed) position are *implicitly* unioned. That's a special-cased convention for exactly one shape (a single generic parameter position), not a general grouping mechanism, and it doesn't help once you're inside a fixed/positional construct like a tuple or an intersection conjunct.

### Options

1. Add an explicit grouping notation that means "these are alternatives" regardless of surrounding context - something that doesn't collide with existing fixed-tuple-parameter syntax (`Array(A, B)`) or the top-level comma-as-union convention.
2. Accept the gap and document it: unions can only appear "shallow" (as the top-level type, or in positions that already have an implicit-union convention like generic params) - anything deeper (tuple slots, intersection conjuncts, etc.) requires going around the string grammar entirely, e.g. by translating from a format that already has real grouping (RBS's `#: ` inline signature syntax, which uses `[...]` for generics/tuples and leaves `()` free for grouping, has no such collision).

Solargraph currently does (2) as a practical matter - for RBS-to-tag translation specifically, we build compound types directly from the parsed RBS AST rather than round-tripping through a flattened tag string, which sidesteps the problem for that one path without solving it for hand-written tags.

apiology:
## Summary

Adds three related type-tag syntax elements:

- **`&` (intersection, closes #1644)**: `Foo & Bar` means a value must
  satisfy both `Foo` and `Bar`, matching Solargraph's syntax
  (castwide/solargraph#1231). Legal in every position a type can appear,
  and always binds tighter than the union or slot separator around it,
  matching RBS's documented precedence. Renders as "both a Foo and a
  Bar" (or "all of a Foo, a Bar, and a Baz" for 3+), to avoid reading
  like two separate values.
- **`|` (union, closes #1699)**: marks a union - a value matching any of
  the listed types. Some type lists already mean a union without it (the
  top level, a hash's key/value lists, `[...]`, and
  `Array<...>`/`Set<...>`), so `,` and `|` land on the same result there.
  Elsewhere, each comma-separated item is a distinct, positional type
  parameter instead (a fixed-order list like `Array(...)`, or `<...>` for
  a name other than `Array`/`Set`) - there, `|` groups alternatives
  within a single one of them: `Array(Foo | Bar, Baz)` is a 2-element
  Array whose first element is a Foo or a Bar, and
  `Result<Success | Failure, Other>` is a Result whose first type
  parameter is a Success or a Failure.
- **`[...]` (closes #1699)**: used the same way parentheses are in
  algebra, to override the default order of operations - e.g. to use a
  union as one conjunct of an intersection, which otherwise has no way to
  mark where the union ends: `[Foo | Bar] & Baz`.

Also documents three pre-existing but previously undocumented anonymous
shorthand forms - `<A>`, `(A)`, `{A=>B}` - where the leading type name can
be omitted and defaults to `Array`/`Hash` (see #1701), and stops
`Foo<A, B>` from always being read as a union: `<...>`'s type parameters
are conventionally used both ways - a homogeneous collection's implicit
union of element type(s) (`Array<String, Symbol>`), or a class's
distinct, positional type parameters (`Result<Success, Failure>`).
`Array`/`Set` (and any name with a single type parameter) keep the union
reading; `Hash<K, V>` gets its own dedicated key/value rendering matching
`Hash{K=>V}`; anything else with 2+ parameters reads neutrally ("a Result
with type parameters (a Success, a Failure)"). This choice is made
entirely at render time (`CollectionType#to_s`) - the parser always
treats `<...>` the same way it already treats `(...)`, with no
name-specific knowledge at all.

Full rules and examples are in the new "Operator Precedence" and
"Overriding the Order of Operations" sections of `docs/Tags.md`, and the
rewritten "Parameterized Types"/"Union Operator" sections.

## Test plan

- [x] `bundle exec rspec spec/tags/types_explainer_spec.rb` - specs for
  `IntersectionType`/`GroupType`/`CollectionType#to_s`, parser-level
  precedence/error cases, and end-to-end `.explain` examples.
- [x] `bundle exec rspec` - full suite green (2830 examples, 0 failures).

### DIFF
--- a/docs/Tags.md
+++ b/docs/Tags.md
@@ -169,18 +169,29 @@ Note that one extra type that is accepted by convention is the `Boolean` type,
 which represents both the `TrueClass` and `FalseClass` types. This type does not
 exist in Ruby, however.
 
-#### Parametrized Types
+#### Parameterized Types
 
 In addition to basic types (like String or Array), YARD conventions allow for
-a "generics" like syntax to specify container objects or other parametrized types.
+a "generics" like syntax to specify container objects or other parameterized types.
 The syntax is `Type<SubType, OtherSubType, ...>`. For instance, an Array might
 contain only String objects, in which case the type specification would be
-`Array<String>`. Multiple parametrized types can be listed, separated by commas.
+`Array<String>`. Multiple type parameters can be listed, separated by commas:
+`Array<String, Fixnum>` can contain any amount of Strings or Fixnums, in any
+order - `Array` and `Set` both treat their type parameters as an implicit
+union, meaning "any of these".
 
-Note that parametrized types are typically not order-dependent, in other words,
-a list of parametrized types can occur in any order inside of a type. An array
-specified as `Array<String, Fixnum>` can contain any amount of Strings or Fixnums,
-in any order. When the order matters, use "order-dependent lists", described below.
+The type name before `<...>` can be omitted, in which case it defaults to
+`Array`: `<String, Fixnum>` means the same thing as `Array<String, Fixnum>`.
+
+Not every type uses its parameters this way, though: `Result<Success, Failure>`
+uses `Success` and `Failure` as two distinct type parameters, not "either of
+these". A type with a single parameter is unambiguous either way. For a name
+that isn't specifically known to mean a union, YARD describes its
+parameters neutrally, without asserting either reading:
+`Result<Success, Failure>` reads as "a Result with type parameters (a
+Success, a Failure)". `Hash<KeyType, ValueType>` is a special case with its
+own dedicated positional rendering (slot 1 is the key type, slot 2 is the
+value type), matching the `Hash{KeyType=>ValueType}` syntax described below.
 
 #### Duck-Types
 
@@ -194,9 +205,24 @@ that responds to the "read" method:
     # @param io [#read] the input object to read from
     def read(io) io.read end
 
+#### Intersection Types
+
+Types joined with `&` describe an intersection: a value that satisfies every
+type listed, rather than any one of them (which is what a union - the `|`
+operator - already means). For instance, an argument that must both
+inherit from `Foo` and respond to `#bar` would be listed as `Foo & #bar`.
+
+    # Accepts any Comparable string.
+    # @param value [String & Comparable] the value to accept
+    def accept(value) end
+
+`&` is legal in every position a type can appear, and always binds tighter
+than any union or slot separator around it - see
+[Operator Precedence](#Operator_Precedence) below.
+
 #### Hashes
 
-Hashes can be specified either via the parametrized type discussed above,
+Hashes can be specified either via the parameterized type discussed above,
 in the form `Hash<KeyType, ValueType>`, or using the hash specific syntax:
 `Hash{KeyTypes=>ValueTypes}`. In the latter case, KeyTypes or ValueTypes can
 also be a list of types separated by commas.
@@ -222,12 +248,73 @@ Keys in the hash-specific syntax are commonly [literal values](#Literals) such
 as symbols (`:key`) or strings (`'key'`, `"key"`), but any type listed in the
 [type conventions](#Type_List_Conventions) is allowed.
 
+The type name before `{...}` can be omitted, in which case it defaults to
+`Hash`: `{K=>V}` means the same thing as `Hash{K=>V}`.
+
 #### Order-Dependent Lists
 
 An order dependent list is a set of types surrounded by "()" and separated by
 commas. This list must contain exactly those types in exactly the order specified.
 For instance, an Array containing a String, Fixnum and Hash in that order (and
 having exactly those 3 elements) would be listed as: `Array(String, Fixnum, Hash)`.
+
+The type name before `(...)` can be omitted, in which case it defaults to
+`Array`: `(String, Fixnum, Hash)` means the same thing as
+`Array(String, Fixnum, Hash)`.
+
+A single slot can itself be a union: `Array(Integer | String, Symbol)` is a
+2-element Array whose first element is an Integer or a String - see
+[Union Operator](#Union_Operator) below.
+
+#### Union Operator
+
+`|` marks a union: a value matching any of the listed types.
+`Integer | String` means an Integer or a String.
+
+Some type lists already mean a union without `|` - a plain comma-separated
+list at the top level, a hash's key or value list, inside `[...]`
+(described below), and inside `Array<...>`/`Set<...>`. In those places `,`
+and `|` come to the same thing, so use whichever reads better:
+`Integer, String` and `Integer | String` describe the same type.
+
+Elsewhere, each comma-separated item is a distinct, positional type
+parameter instead - a fixed-order list like `Array(...)`, or `<...>` for a
+type other than `Array`/`Set` (see
+[Parameterized Types](#Parameterized_Types) above). There, use `|` within a
+single item to say it can be any of several types:
+`Array(Integer | String, Symbol)` is a 2-element Array whose first element
+is an Integer or a String, and `Result<Success | Failure, Other>` is a
+Result whose first type parameter is a Success or a Failure.
+
+#### Operator Precedence
+
+`&`, `,`, and `|` can all appear in the same type, and `&` always binds
+tighter than the union or slot separator around it: `Foo & Bar, Baz` means
+either both a Foo and a Bar, or a Baz - not `Foo`, unioned with `Bar & Baz`.
+
+#### Overriding the Order of Operations
+
+Square brackets `[...]` are used the same way parentheses are in algebra:
+to override the order of operations described above. Types inside `[...]`
+are combined first; hence `[Foo | Bar] & Baz` describes a value that's
+either a Foo or a Bar, and also a Baz - not `Foo`, unioned with
+`Bar & Baz`. Without the brackets, `&` binds tighter than the union
+around it, so `Foo | Bar & Baz` means the latter.
+
+Inside `[...]`, `,` and `|` both mean a union, so `[Foo, Bar]` and
+`[Foo | Bar]` describe the same type.
+
+<p class="note">
+  While you can treat them the same in practice, this <code>[...]</code> is
+  not the same thing as the <code>[Types]</code> brackets that delimit a
+  tag's whole <a href="#Types_Specifier_List">types specifier list</a> -
+  that outer bracket is tag punctuation, not part of any individual type.
+</p>
+
+`[...]` never takes a preceding type name - `[Integer | String]` alone
+just means "an Integer or a String," identical in meaning to the plain
+top-level list `Integer, String`, just usable in more places. It can nest
+inside itself (`[[Foo | Bar] | Baz]`).
 
 #### Literals
 
--- a/lib/yard/tags/types_explainer.rb
+++ b/lib/yard/tags/types_explainer.rb
@@ -35,7 +35,7 @@ module YARD
 
         def to_s(singular = true)
           if name[0, 1] =~ /[A-Z]/
-            singular ? "a#{name[0, 1] =~ /[aeiou]/i ? 'n' : ''} " + name : "#{name}#{name[-1, 1] =~ /[A-Z]/ ? "'" : ''}s"
+            singular ? "#{indefinite_article(name)} " + name : "#{name}#{name[-1, 1] =~ /[A-Z]/ ? "'" : ''}s"
           else
             name
           end
@@ -52,6 +52,11 @@ module YARD
             index += 1
             acc
           end
+        end
+
+        # @return ["a", "an"] whichever indefinite article precedes `word`
+        def indefinite_article(word)
+          word[0, 1] =~ /[aeiou]/i ? "an" : "a"
         end
       end
 
@@ -70,8 +75,71 @@ module YARD
       end
 
       # @private
+      class IntersectionType < Type
+        attr_accessor :types
+
+        def initialize(types)
+          @types = types
+        end
+
+        def to_s(singular = true)
+          # A leading "both"/"all of" disambiguates a single value
+          # satisfying every listed type from what "a Foo and a Bar" alone
+          # could otherwise read as - two separate things.
+          prefix = types.size == 2 ? "both " : "all of "
+          prefix + list_join(types.map {|t| t.to_s(singular) }, with: "and")
+        end
+      end
+
+      # @private
+      class GroupType < Type
+        attr_accessor :types
+
+        def initialize(types)
+          @types = types
+        end
+
+        def to_s(singular = true)
+          "(" + list_join(types.map {|t| disambiguate(t, singular) }) + ")"
+        end
+
+        private
+
+        # {IntersectionType} and a multi-method {DuckType} (`#foo & #bar`)
+        # both render as a bare "X and Y", with no punctuation of their own
+        # to mark where they end. Sitting next to a sibling in this group's
+        # own "or"-joined list, that reads ambiguously (e.g. "a Foo and a
+        # Bar or a Baz" doesn't show whether the "and" or the "or" binds
+        # tighter) even though the parse itself is unambiguous. Wrap those
+        # two cases in their own parens so the group's members are visually
+        # self-delimiting; every other {Type} already is (a bare name, or
+        # something with its own bracketing like {GroupType} itself).
+        def disambiguate(type, singular)
+          rendered = type.to_s(singular)
+          needs_parens = type.is_a?(IntersectionType) ||
+                         (type.is_a?(DuckType) && type.name.include?('&'))
+          needs_parens ? "(#{rendered})" : rendered
+        end
+      end
+
+      # @private
+      #
+      # `<...>`'s type parameters are conventionally used two ways - as a
+      # homogeneous collection's implicit union of element type(s), or as
+      # a class's distinct, positional type parameters, e.g.
+      # `Result<Success, Failure>` - and this class renders either way,
+      # chosen per instance by {#union?}. This keeps that choice entirely
+      # out of the parser: every `<...>` becomes one of these, regardless
+      # of name.
       class CollectionType < Type
         attr_accessor :types
+
+        # Type names known in advance to be genuinely homogeneous
+        # collections, where `<...>`'s type parameters really do mean
+        # "any one of these". Fixed and not user-configurable - YARD has
+        # no syntax for a class to declare its own `<...>` convention, so
+        # this can only ever be a hardcoded, conservative list.
+        UNION_COLLECTION_NAMES = %w[Array Set].freeze
 
         def initialize(name, types)
           @name = name
@@ -79,14 +147,39 @@ module YARD
         end
 
         def to_s(_singular = true)
-          "a#{name[0, 1] =~ /[aeiou]/i ? 'n' : ''} #{name} of (" + list_join(types.map {|t| t.to_s(false) }) + ")"
+          if union?
+            "#{indefinite_article(name)} #{name} of (" + list_join(flattened_types.map {|t| t.to_s(false) }) + ")"
+          else
+            "#{indefinite_article(name)} #{name} with type parameters (" +
+              types.map {|t| t.to_s(true) }.join(", ") + ")"
+          end
+        end
+
+        private
+
+        # A single type parameter is never ambiguous (there's nothing to
+        # distinguish a union from a positional type parameter when
+        # there's only one), so it's read as a union regardless of name;
+        # otherwise, only the known homogeneous-collection names are.
+        def union?
+          types.size <= 1 || UNION_COLLECTION_NAMES.include?(name)
+        end
+
+        # A parsed type parameter that's itself a union (from `Foo | Bar`
+        # in e.g. `Array<Foo | Bar, Baz>`) is a {GroupType}. Since this
+        # whole parameter list is already read as one flat union, absorb
+        # such a parameter's own members into that same list instead of
+        # rendering it as a nested, parenthesized sub-union - the two mean
+        # the same thing here.
+        def flattened_types
+          types.flat_map {|t| t.is_a?(GroupType) ? t.types : [t] }
         end
       end
 
       # @private
       class FixedCollectionType < CollectionType
         def to_s(_singular = true)
-          "a#{name[0, 1] =~ /[aeiou]/i ? 'n' : ''} #{name} containing (" + types.map(&:to_s).join(" followed by ") + ")"
+          "#{indefinite_article(name)} #{name} containing (" + types.map(&:to_s).join(" followed by ") + ")"
         end
       end
 
@@ -134,9 +227,9 @@ module YARD
         end
 
         def to_s(_singular = true)
-          return "a#{name[0, 1] =~ /[aeiou]/i ? 'n' : ''} #{name}" if @key_value_pairs.empty?
+          return "#{indefinite_article(name)} #{name}" if @key_value_pairs.empty?
 
-          result = "a#{name[0, 1] =~ /[aeiou]/i ? 'n' : ''} #{name} with "
+          result = "#{indefinite_article(name)} #{name} with "
           parts = @key_value_pairs.map do |keys, values|
             "keys made of (" + list_join(keys.map {|t| t.to_s(false) }) +
             ") and values of (" + list_join(values.map {|t| t.to_s(false) }) + ")"
@@ -154,9 +247,13 @@ module YARD
           :collection_end => />/,
           :fixed_collection_start => /\(/,
           :fixed_collection_end => /\)/,
+          :group_start => /\[/,
+          :group_end => /\]/,
           :type_name => /#{ISEP}#{METHODNAMEMATCH}|#{NAMESPACEMATCH}|#{LITERALMATCH}|\w+/,
           :symbol => /:#{METHODNAMEMATCH}/,
           :type_next => /[,]/,
+          :intersect => /&/,
+          :union => /\|/,
           :whitespace => /\s+/,
           :hash_collection_start => /\{/,
           :hash_collection_value => /=>/,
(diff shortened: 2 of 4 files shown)


docs/Tags.md
```diff
@@ -222,13 +261,85 @@ Keys in the hash-specific syntax are commonly [literal values](#Literals) such
...
+`Array`: `(String, Fixnum, Hash)` means the same thing as
+`Array(String, Fixnum, Hash)`.
+
+A single slot can itself be a union of types joined with `|`: in
+`Array(Integer | String, Symbol)`, the first slot is an `Integer` or a
+`String`, followed by a `Symbol`. `,` can't be reused for this the way it
+is everywhere else, since inside `(...)` it already means "next slot," not
+"either of these" - see [Grouped Unions](#Grouped_Unions) below.
+
+#### Grouped Unions
```
apiology: Let's document groups and unions separately.  Maybe unions first, then groups could have an example disambiguating the union.

lib/yard/tags/types_explainer.rb
```diff
@@ -154,9 +221,13 @@ class Parser
           :collection_end => />/,
           :fixed_collection_start => /\(/,
           :fixed_collection_end => /\)/,
+          :group_start => /\[/,
+          :group_end => /\]/,
           :type_name => /#{ISEP}#{METHODNAMEMATCH}|#{NAMESPACEMATCH}|#{LITERALMATCH}|\w+/,
           :symbol => /:#{METHODNAMEMATCH}/,
           :type_next => /[,]/,
+          :intersect => /&/,
+          :union_sep => /\|/,
```
apiology: Why _sep here?

docs/Tags.md
```diff
@@ -182,6 +182,26 @@ a list of parametrized types can occur in any order inside of a type. An array
 specified as `Array<String, Fixnum>` can contain any amount of Strings or Fixnums,
 in any order. When the order matters, use "order-dependent lists", described below.
```
apiology: Lets look at this existing paragraph please; it seems to contradict the new information.

docs/Tags.md
```diff
@@ -182,6 +182,26 @@ a list of parametrized types can occur in any order inside of a type. An array
...
+`Set` are treated as the first (implicit-union) kind; any other name with
+two or more parameters is described neutrally, without asserting either
+reading, e.g. `Result<Success, Failure>` reads as "a Result with type
+parameters (a Success, a Failure)". A single parameter is never ambiguous
+(there's nothing to distinguish a union from a positional role when there's
+only one), so it's unaffected either way. `Hash<KeyType, ValueType>` is a
+special case with its own dedicated positional meaning (slot 1 is the key
+type, slot 2 is the value type), matching the `Hash{KeyType=>ValueType}`
+syntax described below - not an implicit union of "KeyTypes or ValueTypes".
+
```
apiology, line 195: I'd think of this more that a given type may treat its argument impllcitly as a union or not, with Array and Set as examples.  

YARD is not a typechecker; the only limited case where it matters to YARD is that type explainer, so it's none of YARD's business whether a given type treats it one way or the other; you can move the hand-wringing over not having syntax to declare types as implicitly union over to TypesExplainer's source code if you like.  No need to air implementation details and confuse folks here.

docs/Tags.md
```diff
@@ -194,6 +214,25 @@ that responds to the "read" method:
...
+    # @param value [String & Comparable] the value to accept
+    def accept(value) end
+
+`&` binds tighter than `,`, so `Foo & Bar, Baz` means `(Foo & Bar), Baz`: a
+value that is either "both a Foo and a Bar" or "a Baz". Use `&` again to
+intersect the same list on both sides of a `,` if that's what you mean:
+`Foo & Bar, Baz & Qux`. `&` is legal in every position a type can appear -
+top level, inside `<...>`, `(...)`, or `{...}` - and always binds tighter
+than whatever separator surrounds it; see [Operator Precedence](#Operator_Precedence)
+below.
```
apiology: How about we move this paragraph until we introduce the idea of a group later on, and use the actual group syntax - [] instead of () which isn't group syntax

docs/Tags.md
```diff
@@ -222,13 +261,86 @@ Keys in the hash-specific syntax are commonly [literal values](#Literals) such
...
+`&`, `,`, and `|` can all appear in the same type, and each means something
+different depending on where it's used, so here's how they combine:
+
+* `&` always binds tighter than whichever separator surrounds it - `,` or
+  `|` at the top level, or the current slot's `|` inside `(...)`.
+  `Foo & Bar, Baz` means `(Foo & Bar), Baz`, and
+  `Array(Foo & Bar | Baz, Qux)` means `Array((Foo & Bar) | Baz, Qux)` - a
+  2-element Array whose first slot is `(Foo & Bar) or Baz` - not
+  `Foo & (Bar, Baz)` or `Foo & (Bar | Baz)`.
+* `,` and `|` are synonyms everywhere - top level, `<...>`, `{...}`, and
```
apiology: , is not a synonym for | within <> type declarations.  Only some types document their slots as implicitly unioned (e.g., Array<> and Set<> within the standard library)

lib/yard/tags/types_explainer.rb
```diff
@@ -181,34 +259,113 @@ def parse(until_tokens: [:parse_end])
...
+        #
+        # `[...]` groups a union (spelled with either `,` or `|`) into a
+        #   single type usable as one conjunct of an intersection at the
+        #   top level, where a bare union would otherwise just add another
+        #   independent top-level item instead: `[A | B] & C` groups `A`
+        #   and `B` before intersecting with `C`, where `A | B & C` would
+        #   parse as the two top-level items `A` and `B & C`.
+        # @return [Array(Array<Type>, Symbol)] the parsed types and the
+        #   token that ended the list
+        def parse_until(until_tokens, slot_pipe: false)
```
apiology, line 306: Could we make Set<> and Array<> a special case within the text rendering code instead of within the parser itself, or is there a need or advantage for the parser not to just treat these as slots until it needs to describe types?

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Done - split into separate "Union Operator" and "Grouping" sections in ac02df02 / 1dc99166.

lib/yard/tags/types_explainer.rb
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Renamed `:union_sep` to `:union` to match `:intersect`'s naming - no reason for the asymmetry. See 1dc99166.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Rewrote in 1dc99166: the order-independence claim now applies specifically to the implicit-union case (Array/Set); positional `<...>` types like `Result<Success, Failure>` are called out as caring about order, so it no longer contradicts the Parametrized Types material below it.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Rewrote in 1dc99166 along the lines you suggested: describes it as "a given type may treat its parameters as an implicit union or not" and dropped the "YARD has no syntax to declare which..." framing. That reasoning still lives in the Parser source (see the `UNION_COLLECTION_NAMES` and `ParametrizedType` comments), just not in the user-facing docs.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Moved in 1dc99166 - Intersection Types now just states `&` is legal everywhere and points to Operator Precedence, which owns the full worked example. Also swapped the bare `(Foo & Bar)` prose notation for the actual `[Foo & Bar]` group syntax there.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Good catch - fixed in 1dc99166, both in docs and behavior. The Operator Precedence bullet now says `,`/`|` are synonyms in `<...>` only for the implicit-union types (Array/Set); for any other name, `<...>`'s parameters are positional, so `|` doesn't apply there. I made this a real parse-time SyntaxError too (e.g. `Result<Success | Failure>` now fails to parse) rather than just a doc note, since silently accepting `|` somewhere it has no union meaning to apply to seemed worse than rejecting it - `|` always means union, per our earlier agreement, so a context that doesn't implicitly union its slots shouldn't accept it at all.

lib/yard/tags/types_explainer.rb
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Kept it in the parser, for a reason that got stronger after your other comment: rejecting `Result<Success | Failure>` (the new parse-time SyntaxError from the , vs | in <> thread) needs to know at parse time whether a given `<...>` is implicit-union or positional, so it can tell whether the `|` it just saw was meaningful. If the Array/Set decision moved to the renderer, the parser would still need its own copy of that allow-list to know when to raise, and we'd have the same list living in two places able to drift apart. Keeping the single UNION_COLLECTION_NAMES check in the parser is what lets both the type-selection and the syntax check share one source of truth. If the syntax check gets dropped later, moving the classification to rendering would become more attractive.

docs/Tags.md
```diff
@@ -177,10 +177,30 @@ The syntax is `Type<SubType, OtherSubType, ...>`. For instance, an Array might
...
-Note that parametrized types are typically not order-dependent, in other words,
-a list of parametrized types can occur in any order inside of a type. An array
-specified as `Array<String, Fixnum>` can contain any amount of Strings or Fixnums,
-in any order. When the order matters, use "order-dependent lists", described below.
+A given type may treat its `<...>` parameters as an implicit union, meaning
+"any of these" - `Array` and `Set` do, so `Array<String, Fixnum>` can contain
+any amount of Strings or Fixnums, in any order - or as distinct, positional
+roles instead, in which case order matters (for instance,
+`Result<Success, Failure>` - `Success` and `Failure` are not "either of
+these", they're two different roles). Either way, this is still different
```
apiology: Let's break sentence 1 into two sentences to reduce ambiguity.

docs/Tags.md
```diff
@@ -177,10 +177,30 @@ The syntax is `Type<SubType, OtherSubType, ...>`. For instance, an Array might
...
+from an "order-dependent list" (described below), which asserts an exact
+sequence of elements rather than naming parameter roles.
+
+The type name before `<...>` can be omitted, in which case it defaults to
+`Array`: `<String, Fixnum>` means the same thing as `Array<String, Fixnum>`.
+
+Only `Array` and `Set` are treated as the implicit-union kind; any other
+name with two or more parameters is described neutrally, without asserting
```
apiology: Again, from the user's perspective, Array and Set are EXAMPLES of types which use their slots as an implicit unions.  YARD has little to no interest in what types do with their slots at this point; only the type describer cares about knowing the difference, and who knows, maybe someday we'll provide a mechanism to describe the semantics of a given type.

docs/Tags.md
apiology: Please read the above note again in the context of the new comment above.

docs/Tags.md
```diff
@@ -194,6 +214,24 @@ that responds to the "read" method:
...
+type listed, rather than any one of them (which is what a comma-separated
+list, a union, already means). For instance, an argument that must both
+inherit from `Foo` and respond to `#bar` would be listed as `Foo & #bar`.
+
+    # Accepts any Comparable string.
+    # @param value [String & Comparable] the value to accept
+    def accept(value) end
+
+`&` is legal in every position a type can appear - top level, inside
+`<...>`, `(...)`, `{...}`, or `[...]` - and always binds tighter than
```
apiology: Drop the examples here - I think those are implicit and I don't want to forward reference []

docs/Tags.md
```diff
@@ -222,13 +260,92 @@ Keys in the hash-specific syntax are commonly [literal values](#Literals) such
...
+  `|` at the top level, or the current slot's `|` inside `(...)`.
+  `Foo & Bar, Baz` means the same thing as `[Foo & Bar], Baz`, and
+  `Array(Foo & Bar | Baz, Qux)` means the same thing as
+  `Array([Foo & Bar] | Baz, Qux)` - a 2-element Array whose first slot is
+  either both a Foo and a Bar, or a Baz - not `Foo & [Bar, Baz]` or
+  `Foo & [Bar | Baz]`.
+* `,` and `|` are synonyms everywhere a union is legal - top level,
+  `{...}`, `[...]`, and inside `<...>` for the types whose parameters are
+  an implicit union (`Array`/`Set`; see
+  [Parametrized Types](#Parametrized_Types) above) - except directly
```
apiology: Let's fix spelling here and throughout - we're missing an 'e'

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Split into two sentences in 116d84bf.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Reframed in 116d84bf: "Array and Set are examples of types whose parameters are known to be an implicit union" rather than "Array and Set are treated as..." - keeps the framing on what the type does, not on YARD's internal allow-list.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Dropped the bracket-position list in 116d84bf - now just "legal in every position a type can appear" with no forward reference to [...].

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Good catch - fixed "Parametrized" -> "Parameterized" throughout in 116d84bf: the doc heading (pre-existing, from before this PR), the ParameterizedType class name, and all spec references.

docs/Tags.md
```diff
@@ -169,18 +169,39 @@ Note that one extra type that is accepted by convention is the `Boolean` type,
...
+`Array<String>`. Multiple parameterized types can be listed, separated by commas.
+
+A given type may treat its `<...>` parameters as an implicit union, meaning
+"any of these", or as distinct, positional roles instead, in which case
+order matters. For instance, `Array<String, Fixnum>` can contain any
+amount of Strings or Fixnums, in any order, while in
+`Result<Success, Failure>`, `Success` and `Failure` are not "either of
+these" - they're two different roles. Either way, this is still different
+from an "order-dependent list" (described below), which asserts an exact
+sequence of elements rather than naming parameter roles.
```
apiology: Let's drop this last sentence and introduce the concept when we get to it.

apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Dropped in 19361adb.

lib/yard/tags/types_explainer.rb
apiology: Why would the parser reject "Result<Success | Failure>"?  That should be a Result with one slot which is either a Success or Failure.  Or do we have hard-coded knowledge of the number of slots a given type can/must take?

docs/Tags.md
```diff
@@ -169,18 +169,37 @@ Note that one extra type that is accepted by convention is the `Boolean` type,
...
-specified as `Array<String, Fixnum>` can contain any amount of Strings or Fixnums,
-in any order. When the order matters, use "order-dependent lists", described below.
+`Array<String>`. Multiple parameterized types can be listed, separated by commas.
+
+A given type may treat its `<...>` parameters as an implicit union, meaning
+"any of these", or as distinct, positional roles instead, in which case
+order matters. For instance, `Array<String, Fixnum>` can contain any
+amount of Strings or Fixnums, in any order, while in
+`Result<Success, Failure>`, `Success` and `Failure` are not "either of
+these" - they're two different roles.
```
apiology: How about "type parameters"?  I don't want to give people the impression that "role" is a term of art

docs/Tags.md
```diff
@@ -169,18 +169,37 @@ Note that one extra type that is accepted by convention is the `Boolean` type,
...
+
+The type name before `<...>` can be omitted, in which case it defaults to
+`Array`: `<String, Fixnum>` means the same thing as `Array<String, Fixnum>`.
+
+`Array` and `Set` are examples of types whose parameters are known to be
+an implicit union; any other name with two or more parameters is
+described neutrally, without asserting either reading, e.g.
```
apiology: We're not documenting TypesDescriber here; we're telling a user of YARD how they should understand types they see and how they can describe their own types.  Array and Set and the top level types having implicit union is worth about a sentence to the end user; another sentence maybe for the Hash syntax.  The rest of this paragraph is internal documentation about TypesExplainer; move it there if you think it's important.

docs/Tags.md
```diff
@@ -169,18 +169,37 @@ Note that one extra type that is accepted by convention is the `Boolean` type,
...
+`Result<Success, Failure>` reads as "a Result with type parameters (a
+Success, a Failure)". A single parameter is never ambiguous (there's
+nothing to distinguish a union from a positional role when there's only
+one), so it's unaffected either way. `Hash<KeyType, ValueType>` is a
+special case with its own dedicated positional meaning (slot 1 is the key
+type, slot 2 is the value type), matching the `Hash{KeyType=>ValueType}`
+syntax described below - not an implicit union of "KeyTypes or ValueTypes".
+Since `|` always means union, using it inside a `<...>` type whose
+parameters aren't known to be a union is a syntax error rather than being
+silently accepted; use `,` there instead.
```
apiology: This last sentence is simply flat out wrong; we should treat Result<String | Integer, Float> is perfectly fine syntax meaning two slots, the first of which is a union

docs/Tags.md
```diff
@@ -222,13 +258,92 @@ Keys in the hash-specific syntax are commonly [literal values](#Literals) such
...
 
+The type name before `(...)` can be omitted, in which case it defaults to
+`Array`: `(String, Fixnum, Hash)` means the same thing as
+`Array(String, Fixnum, Hash)`.
+
+A single slot can itself be a union of types joined with `|`: in
+`Array(Integer | String, Symbol)`, the first slot is an `Integer` or a
+`String`, followed by a `Symbol`. `,` can't be reused for this the way it
+is everywhere else, since inside `(...)` it already means "next slot," not
+"either of these" - see [Union Operator](#Union_Operator) below.
```
apiology: Seems like a lot of detail when we're trying to get concepts across; thoughts?

docs/Tags.md
```diff
@@ -222,13 +258,92 @@ Keys in the hash-specific syntax are commonly [literal values](#Literals) such
...
+("either of these") almost everywhere: `Integer | String` means the same
+thing as `Integer, String`, and likewise inside `<...>` and `{...}`. The
+two can be mixed freely, and read the same either way.
+
+The one place they diverge is directly inside an order-dependent list's
+`(...)` - see [Order-Dependent Lists](#Order-Dependent_Lists) above. There,
+`,` already means "next slot," so `|` can't be a synonym for it without
+losing its own meaning; instead, `|` keeps meaning "either of these," just
+scoped to a single slot: `Array(Integer | String, Symbol)` is a 2-element
+Array, not a 3-element one.
```
apiology: Rephrase this whole section; union operator is a union; `,` also represents a union in some cases, which you can describe.  

"|" being a synonym for "," is a confusing way to describe it.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Switched "role"/"roles" to "type parameter(s)" throughout, in fba81e29.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Trimmed to two sentences on the user-facing behavior (Array/Set implicit union, Hash's special rendering) in fba81e29 - the rest of the reasoning about why YARD can't tell positional from union in general now lives only in the Parser source comments.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

You're right - reverted the SyntaxError. `Result<Success | Failure, Other>` now parses as two type parameters, the first of which is a union, exactly the way `Array(Success | Failure, Other)` already groups a union within one tuple slot. Implementation-wise: whether a `<...>` name is known to implicitly union its whole list (Array/Set) now just selects `slot_pipe` the same way `(...)` always has, rather than being a separate reject-or-allow check. Fixed in fba81e29.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Cut it down to the equivalence statement and a pointer to Union Operator, dropping the meta-commentary about why ',' can't be reused - in fba81e29.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Rewrote in fba81e29: leads with "'|' marks a union" and explains that some type lists already mean a union (so ',' and '|' land on the same result there), while others treat each comma-separated item as a distinct slot (where '|' groups within one item) - no "synonym" language.

docs/Tags.md
```diff
@@ -222,13 +248,79 @@ Keys in the hash-specific syntax are commonly [literal values](#Literals) such
...
+is an Integer or a String, and `Result<Success | Failure, Other>` is a
+Result whose first type parameter is a Success or a Failure.
+
+#### Grouping
+
+Square brackets, `[...]`, group a list of types - written with either `,`
+or `|` - into a single type that can be used anywhere a type is expected.
+Its main use is grouping a union for use as one conjunct of an
+intersection, since a bare union there has no way to mark where it ends:
+`[Integer | String] & Comparable`.
```
apiology: People understand using algebra and using parentheses for grouping; just describe it like that rather than zeroing in so hard on the fact that right now the precedence between union and intersection is the only reason you'd need to use it.  Describe the broader concept and use that driving example as just an example.

docs/Tags.md
```diff
@@ -222,13 +248,79 @@ Keys in the hash-specific syntax are commonly [literal values](#Literals) such
...
+<p class="note">
+  This <code>[...]</code> is unrelated to the <code>[Types]</code> brackets
+  that delimit a tag's whole <a href="#Types_Specifier_List">types
+  specifier list</a> - that outer bracket is tag punctuation, not part of
+  any individual type. <code>[...]</code> as described here only has
+  meaning <em>inside</em> a type, e.g. as one slot of
+  <code>Array(...)</code> or one conjunct of an intersection.
+</p>
+
+`[...]` never takes a preceding type name and is never itself a collection
```
apiology: "is not itself"

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Reworked in 138a26ca: leads with "just like parentheses in algebra" now, and the union/intersection precedence case is framed as one example rather than the section's sole justification.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Dropped "is never itself a collection" in 138a26ca - that was leaning on the internal CollectionType name, and wasn't saying anything the "never takes a preceding type name" sentence right after it doesn't already cover.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Follow-up, in ad176db6/038f1e73: the "parentheses group a list" framing was itself wrong (parens group an expression, not a comma list), and on reflection "grouping" wasn't the right concept at all - [...] doesn't introduce a new kind of type the way Intersection/Hashes/Order-Dependent Lists do, it's a precedence-override device. Renamed the section to "Overriding the Order of Operations", moved it after Operator Precedence, and rewrote it around the Wikipedia description of algebraic parentheses, dropping "group"/"grouping" entirely.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Superseded by the broader rewrite in 038f1e73 - the whole section was reframed around "overriding order of operations" rather than "grouping", so this sentence doesn't exist in that form anymore.

lib/yard/tags/types_explainer.rb
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

No hardcoded slot-count knowledge - that SyntaxError was wrong and got reverted in fba81e29 (see the parallel thread on the , vs | in <> comment). `Result<Success | Failure>` now parses as exactly what you'd expect: a single type parameter that's itself a union. Confirmed: `YARD::Tags::TypesExplainer.explain("Result<Success | Failure>")` => "a Result of ((Successs or Failures))" - same doubled-parens convention already used when a GroupType lands inside a single Array(...)/Hash{} slot elsewhere in the suite.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Addressed together with the comment above (3696184322) in fba81e29 - reframed as "Array and Set are examples of types whose parameters are known to be an implicit union" rather than a YARD-decision framing.

docs/Tags.md
```diff
@@ -194,9 +205,24 @@ that responds to the "read" method:
     # @param io [#read] the input object to read from
     def read(io) io.read end
 
+#### Intersection Types
+
+Types joined with `&` describe an intersection: a value that satisfies every
+type listed, rather than any one of them (which is what a comma-separated
+list, a union, already means). For instance, an argument that must both
```
apiology: > (which is what a comma-separated list, a union, already means)

(which is what a union - the | operator - already means)

docs/Tags.md
```diff
@@ -222,13 +248,75 @@ Keys in the hash-specific syntax are commonly [literal values](#Literals) such
...
+are combined first; hence `[Foo | Bar] & Baz` describes a value that's
+either a Foo or a Bar, and also a Baz - not `Foo`, unioned with
+`Bar & Baz`. Without the brackets, `&` binds tighter than the union
+around it, so `Foo | Bar & Baz` means the latter.
+
+Inside `[...]`, `,` and `|` both mean a union, so `[Foo, Bar]` and
+`[Foo | Bar]` describe the same type.
+
+<p class="note">
+  This <code>[...]</code> is unrelated to the <code>[Types]</code> brackets
```
apiology: While you can treat them the same in practice, this <code>[...]</code> is not the same thing as the...

docs/Tags.md
```diff
@@ -222,13 +248,75 @@ Keys in the hash-specific syntax are commonly [literal values](#Literals) such
...
+  that delimit a tag's whole <a href="#Types_Specifier_List">types
+  specifier list</a> - that outer bracket is tag punctuation, not part of
+  any individual type. <code>[...]</code> as described here only has
+  meaning <em>inside</em> a type.
```
apiology: Drop the last sentence.  We designed [] type syntax so that to the user they can express it similarly to the [] tag syntax.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Applied verbatim in 4ac09acf.

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Applied in 4ac09acf: "While you can treat them the same in practice, this [...] is not the same thing as the [Types] brackets..."

docs/Tags.md
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Dropped in 4ac09acf.

lib/yard/tags/types_explainer.rb
apiology: Could we make Set<> and Array<> a special case within the text rendering code instead of within the parser itself, or is there a need or advantage for the parser not to just treat these as slots until it needs to describe types?

apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Answering this directly now, with the current (post-reversal) design - my earlier answer's rationale (a SyntaxError needing to know the allow-list) is gone, but the reason to keep it in the parser got stronger, not weaker.

Whether a `<...>` name is allow-listed now selects `slot_pipe` for that nested parse (lib/yard/tags/types_explainer.rb:354: `slot_pipe = is_fixed || !UNION_COLLECTION_NAMES.include?(name)`), and `slot_pipe` controls how `|` combines with `,` *during parsing itself* - whether `Foo | Bar, Baz` becomes one flat 2-item union (Array/Set) or two slots where the first is itself a union (everything else, e.g. Result<Success | Failure, Other>). That's a difference in the shape of the parsed type tree, not just a difference in how an already-built tree gets described in English. By the time rendering runs, the tree for the Array/Set case and the tree for the Result case are already structurally different - there's no single generic tree the renderer could look at afterward and retroactively decide "oh, this comma meant something else." So this specific check can't move to the renderer without changing what gets parsed, not just what gets said about it.

lib/yard/tags/types_explainer.rb
```diff
@@ -83,6 +131,29 @@ def to_s(_singular = true)
...
+      class ParameterizedType < Type
+        attr_accessor :types
+
+        def initialize(name, types)
+          @name = name
+          @types = types
+        end
+
+        def to_s(_singular = true)
+          "a#{name[0, 1] =~ /[aeiou]/i ? 'n' : ''} #{name} with type parameters (" +
```
apiology: Let's extract out the a vs an to a method please

lib/yard/tags/types_explainer.rb
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Implemented in e5643ea9. The parser now always treats `<...>` the way `(...)` already did - `,` separates type parameters, `|` groups within one of them - with no name lookup at all. `UNION_COLLECTION_NAMES` only decides which Type subclass renders the resulting parameter list, right where that subclass gets picked.

The one adjustment needed: CollectionType (Array/Set) now flattens a GroupType parameter (from e.g. `Array<Foo | Bar, Baz>`) into its own union list when rendering, rather than nesting it in parens, since the whole parameter list is already read as a union there. Output is identical to before for every previously-documented case - verified via spec (two tests that asserted the pre-flattening internal tree shape now assert the rendered string instead, which is the actual guarantee). Full suite: 2830 examples, 0 failures.

lib/yard/tags/types_explainer.rb
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Extracted into a protected Type#indefinite_article(word) method, in 1471edb0 - used by all 5 call sites (Type, CollectionType, ParameterizedType, FixedCollectionType, HashCollectionType).

lib/yard/tags/types_explainer.rb
apiology: 🤖 *Posted by Claude, not Vince — acting on his behalf via his GitHub credentials.*

Taken further in d78cc24e, per your follow-up: UNION_COLLECTION_NAMES no longer appears anywhere in the Parser class. Merged ParameterizedType into CollectionType - the parser now always builds a CollectionType for <...> (aside from the pre-existing Hash<K,V> and fixed-tuple cases), and CollectionType#to_s decides, per-instance via a private #union? check, whether to render as a flat union or neutrally. The constant now lives only on CollectionType, read only at render time. Output unchanged for every case; full suite 2830 examples, 0 failures.

apiology:
@lsegal this one is ready for review

Issue #1701, opened 2026-07-31
Anonymous collection/hash shorthand (<A>, (A), {A=>B}) documented at yardoc.org/types but not in docs/Tags.md


apiology:
*Posted by Claude, not the account owner - acting on their behalf via their GitHub credentials.*

`<A>`, `(A)`, and `{A=>B}` (no preceding class/type name) are all valid, working shorthand: the name defaults to `Array` for the first two and `Hash` for the third. This is:

- Implemented in `YARD::Tags::TypesExplainer::Parser` (`name ||= "Array"` / `name ||= "Hash"` in `lib/yard/tags/types_explainer.rb`).
- Covered by existing specs: `"allows a collection type without a name"`, `"allows a fixed collection type without a name"`, and `"allows a hash collection type without a name"` in `spec/tags/types_explainer_spec.rb`.
- Documented externally at https://yardoc.org/types.html, which lists `<String, Symbol>`, `(String, Symbol)`, and `{Key=>Value}` together as "shorthands for collections and hashes."

But `docs/Tags.md` - which is what ships in the gem and is linked to from the project's own README/docs - has no mention of this at all. Its "Parametrized Types" section only shows the named form (`Array<String>`), "Order-Dependent Lists" only shows the named form (`Array(String, Fixnum, Hash)`), and "Hashes" only shows the named form (`Hash{KeyTypes=>ValueTypes}`). Someone reading only `docs/Tags.md` (not the separate yardoc.org/types page) would have no way to discover that omitting the name is valid.

Filing this since it came up while working on #1644 / #1699 - noticed the behavior is real (verified against the actual parser and its specs) but the local docs don't mention it, only the external page does.

## Suggested fix

Add a line to each of the three relevant `docs/Tags.md` sections (Parametrized Types, Order-Dependent Lists, Hashes) noting that the leading type/class name can be omitted, in which case it defaults to `Array` (or `Hash`), matching what https://yardoc.org/types.html already documents.

lekemula:
TIL!

PR #1702, opened 2026-07-31
"[A | B]" grouped-union syntax for type tags

### LINKED ISSUE #1699: No general grouping syntax for unions nested inside tuples, intersections, etc.

apiology:
## Summary

Closes #1699. There was no way to nest a union inside a position where `,`
already means something else - most concretely, an order-dependent list's
positional slots (`Array(A, B)` already means "2 slots", so `,` can't also
mean "alternative for this slot" there).

`[...]` is a new, dedicated grouping construct - unlike `<...>`, `(...)`,
and `{...}`, it never takes a preceding type name and never means a
collection; `[A | B]` alone just means "A or B", exactly like the plain
top-level list `A, B`, but usable anywhere a type is expected. `|` is only
meaningful inside `[...]`, and `,` is not allowed inside `[...]` - the two
never compete for the same position, so there's nothing to disambiguate
(no need for a "cannot mix `,` and `|`" rule, or extra nested parens).

This resolves the motivating example from the issue directly:
`Array([Integer | String], Symbol)` is a 2-element tuple whose first
element is an `Integer` or a `String`, followed by a `Symbol` - no extra
nesting required.

## Why `[]` and not `()` or reusing `,`

- `()` already has an established meaning (`Array(A, B)`, an
  order-dependent tuple) that this repo isn't changing here, to leave that
  syntax free for a possible future use.
- `[` has no meaning at all today in `YARD::Tags::TypesExplainer::Parser`
  (only `<`, `(`, `{` are wired up as bracket-openers there), even though
  the outer tag-value tokenizer (`DefaultFactory::TYPELIST_OPENING_CHARS`)
  already balances it generically - so it's free to claim without any
  parsing-level conflict.
- Reusing the existing `,`-for-union convention *inside* the grouping
  brackets was considered and rejected: it would make `[A, B]` and
  `[A | B]` two spellings of the same thing, and (worse) invites
  misreading `[A, B]` as an array/tuple literal by analogy with other
  ecosystems, rather than as a plain 2-way alternative. `|` avoids both.

## What else changed

Also documents three pre-existing (but previously undocumented in
`docs/Tags.md`) anonymous shorthand forms - `<A>`, `(A)`, and `{A=>B}` -
where the leading type name can be omitted and defaults to `Array` or
`Hash` respectively. These were already implemented and spec-tested, and
documented externally at https://yardoc.org/types.html, but absent from
this repo's own docs; see #1701.

## Test plan

- [x] `bundle exec rspec spec/tags/types_explainer_spec.rb` - added specs
  for `GroupType#to_s`, parser-level grouping/error cases, and end-to-end
  `.explain` examples.
- [x] `bundle exec rspec` - full suite green (2802 examples, 0 failures).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # main
 
+- Add `[A | B]` grouped-union syntax for type tags (closes #1699); document the
+  existing anonymous `<A>`, `(A)`, and `{A=>B}` shorthand forms
 - Fix duplicate "View source" links after client-side navigation in default HTML template
 
 # [0.9.45] - July 14th, 2026
--- a/docs/Tags.md
+++ b/docs/Tags.md
@@ -182,6 +182,9 @@ a list of parametrized types can occur in any order inside of a type. An array
 specified as `Array<String, Fixnum>` can contain any amount of Strings or Fixnums,
 in any order. When the order matters, use "order-dependent lists", described below.
 
+The type name before `<...>` can be omitted, in which case it defaults to
+`Array`: `<String, Fixnum>` means the same thing as `Array<String, Fixnum>`.
+
 #### Duck-Types
 
 Duck-types are allowed in type specifier lists, and are identified by method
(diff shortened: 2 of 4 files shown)


apiology:
Superseded by https://github.com/lsegal/yard/pull/1700, which now combines this PR's grouped-union work with the intersection-type PR into a single PR, per review feedback. Closing this one.

PR #1706, opened 2026-08-22
html: don't resolve links within code block

0x1eef:
# Description

This patch fixes a bug where a markdown file parsed by yardoc 
(with the redcarpet provider) would emit a warning because yard 
tries to parse Ruby code for doc references (eg "see {Foo}").

The bug requires that `<details><summary>..</summary></details>`
is used alongside a GitHub-flavored markdown codeblock. The 
redcarpet renderer might also be relevant, and the new test 
depends on it.

The diff is a bit noisy because I moved the previous `resolve_links` 
implementation into a private method. The actual new changes in 
lib/ are probably 3 lines, though.

# Completed Tasks

### DIFF
--- a/lib/yard/templates/helpers/html_helper.rb
+++ b/lib/yard/templates/helpers/html_helper.rb
@@ -233,47 +233,13 @@ module YARD
       # @param [String] text the text to resolve links in
       # @return [String] HTML with linkified references
       def resolve_links(text)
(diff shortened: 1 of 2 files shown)


0x1eef:
I am having some trouble with Ruby2 syntax. Working on it.

